PR #29, opened 2020-09-14
Change license from LGPL2.1+ to CC0

MSoegtropIMC:
This commit changes the license from LGPL2.1+ to CC0.

Me and @Zimmi48 are the sole contributors of this project. @Zimmi48 : please write below a comment that you agree to this.

The following shell scripts and batch files are based on files in the coq project: 
```
coq/dev/build/windows/makecoq_mingw.sh -> coq-platform/coq_platform_make.sh
coq/dev/build/windows/MakeCoq_MinGW.bat -> coq-platform/coq_platform_make_windows.bat
coq/dev/build/windows/configure_profile.sh -> coq-platform/windows/configure_profile.sh
```
authored by me during my time at Intel and have been published under LGPL2.1+ by me and Intel. I do this change without written agreement from Intel for these reasons:

- From the original file `coq/dev/build/windows/makecoq_mingw.sh` only these non comment lines remain after the move to opam:
```
set -o nounset
set -o errexit
else
fi
}
  fi
}
}
  done
}
}
  fi
}
  fi
  done
fi
```
The other two files only contain only simple parameter parsing and configuration instructions which are factual consequences of requirements and thus not result of a creative act and cannot be copyrighted in their own right. The copyright was only valid as part of the larger work which is now void.
- I did most of the work on the Coq Windows Installer in my free time - the publication by me *and* Intel was done for clarification that Intel agrees to the publication of an employee. It should be noted that files published as open source by Intel usually do not have an author attribution, while the discussed files are attributed to me as author.
- I also drove the discussion on which license to choose at Intel. The only reason for choosing LGPL2.1+ was because this work has been contributed to the Coq project which is LGPL2.1+. The sole consideration was convenience for the project we contributed to and its users. Since the functionality has been moved to opam and separated from the core Coq project, and since it contains a lot of opam repo patches and since the used opam repositories are CC0, it is better for the project to further distribute this as CC0. I have no doubt that Intel would follow this decision. I did not ask because the question lacks substance and the business value of this is in negligible compared to the cost for discussing this.

Each of the above reasons alone is sufficient to justify the change.


Zimmi48 (approved):
I approve this change of license.

PR #43, opened 2020-12-13
Coq platform on Mac: revisions

Blaisorblade:
- Nitpicks while reviewing.
- IMHO, it's easier to state the misbehavior without blaming Homebrew, if the problem is not fully diagnosed.


MSoegtropIMC:
Thanks for the review! It is indeed better to stick to the pure facts.

PR #50, opened 2020-12-20
[ci] Add dmg package

gares:
TODO:
- [x] generate readme at the root of the DMG
  - [x] turn license fields which begin with http:// into hyperlinks (see VST opam license field)
  - [x] menhir lacks a license field (inserted "unknown - please check <homepage>")
  - [x] bonus: if `https://spdx.org/licenses/${plicense}.html` exists, make an hyperlink (e.g. use `if curl --head --silent --fail https://spdx.org/licenses/${plicense}.html >/dev/null; then ...`)
  - [x] add a link to the homepage for clarification in case the license is not spdx
  - [x] properly escape description for HTML
  - [x] add explicit note on non-commercial software if compcert is included
  - [x] give a less verbose listing of secondary packages (no description)
  - [ ] add link to MacPorts and/or Homebrew package for system dependencies (the opam version information is useless for many conf packages)
- [ ] move .plist file here from Coq
- [ ] share whitelist among windows/osx (and snap, if possible)
- [x] move `${APP_ABSDIR}/Contents/MacOS/coqide` shell to `bin/`, leave there a symlink to bin, rename coqide to coqide.exe
- [x] fix coqide shell script to export GTK_IM_MODULE_FILE
- [x] fix coqide shell script to export GDK_PIXBUF_MODULE_FILE
  - [x] create a loaders.cache file with the paths and ship it (I could not find a better way than putting absolute paths here)
  - [x] test it (it works, but uses absolute paths)
- [x] make add_files_using_macports_package work also on brew
  - [x] it builds
  - [x] it actually works.
- [x] separate win/ubuntu/osx CI workflows, since the space for artifacts is shared (name collision), and one can only start/stop one entire workflow, moreover artifacts only show up for download when the slowest finishes, and finally we only share the cron and platform args part of the file (3 lines)
  - [x] let one start workflows manually passing custom options to the platform script
- [ ] add a CI job using matrix which installs macports (by default brew is installed, not ports)
- [x] allow files which have been modified since opam install

### DIFF
--- a/macos/create_installer_macos.sh
+++ b/macos/create_installer_macos.sh
@@ -9,7 +9,7 @@
 ###################### CREATE MAC DMG INSTALLER ######################
 
 # Options:
-# -quick|-q   : disable ZIP compression of DMG file (much faster to create and install for tests)
+# -quick|-q   : disable BZIP compression of DMG file (much faster to create and install for tests)
 # -install|-i : install package to /Application after creating it
 # -otooldump  : provide the output of otool -L for all executables
 
@@ -24,14 +24,14 @@ set -o errexit
 
 ###### Parse command line ######
 
-ZIPCOMPR=9
+ZIPCOMPR="-format UDBZ" # bzip
 INSTALL='N'
 OTOOLDUMP='N'
 
 for arg in "$@"
 do
   case "${arg}" in
-    -quick|-q) ZIPCOMPR=0 ;;
+    -quick|-q) ZIPCOMPR="-format UDRO" ;;
     -install|-i) INSTALL='Y' ;;
     -otooldump) OTOOLDUMP='Y' ;;
     *) echo "ERROR: Unknown command line argument ${arg}!"; false;;
@@ -39,13 +39,6 @@ do
 done
 
 ###### Check if required system utilities are installed #####
-
-if ! command -v port  &> /dev/null
-then
-  echo "This script assumes that system dependencies have been installed via MacPorts."
-  echo "If you are using Homebrew, please adjust add_files_using_macports_package."
-  exit 1
-fi
 
 command -v gfind &> /dev/null || ( echo "Install gfind (eg. sudo port install findutils)" ; exit 1)
 command -v grealpath &> /dev/null || ( echo "Install gfind (eg. sudo port install coreutils)" ; exit 1)
@@ -64,7 +57,11 @@ mkdir logs
 coqpackagefull=$(opam list --installed-roots --short --columns=name,version coq | sed 's/ /./')
 opam source --dir=coq/ ${coqpackagefull}
 
-##### Get the version of Coq #####
+##### Get the version of Coq and the Platform #####
+
+source ../coq_platform_switch_name.sh
+
+echo "##### Coq platform version = ${COQ_PLATFORM_VERSION} #####" 
 
 COQ_VERSION=$(coqc --print-version | cut -d ' ' -f 1)
 
@@ -77,8 +74,8 @@ echo "##### Coq version = ${COQ_VERSION} (Mac app version=${COQ_VERSION_MACOS}) 
 
 # Folder and image names
 
-APP_NAME="Coq_${COQ_VERSION}.app"
-DMG_NAME="coq-${COQ_VERSION}-installer-macos"
+APP_NAME="Coq_Platform_${COQ_PLATFORM_VERSION}.app"
+DMG_NAME="coq-platform-${COQ_PLATFORM_VERSION}-installer-macos"
 APP_ABSDIR="_dmg/${APP_NAME}"
 RSRC_ABSDIR="${APP_ABSDIR}/Contents/Resources"
 BIN_ABSDIR="$RSRC_ABSDIR/bin"
@@ -97,12 +94,22 @@ mkdir -p ${DYNLIB_ABSDIR}           # System shared libraries
 # The opam prefix - stripped from absolute paths to create relative paths
 OPAM_PREFIX="$(opam conf var prefix)"
 
-##### MacPorts folder variables #####
+##### MacPorts/Brew folder variables #####
 
-# If someone knows a better way to find out where port is installed, please let me know!
-
+set +e
 PORTCMD="$(which port)"
-PORTDIR="${PORTCMD%bin/port}"
+set -e
+
+if [ -z "${PORTCMD}" ]; then
+  PKG_MANAGER=brew
+  PKG_MANAGER_ROOT="/usr/local/"
+  PKG_MANAGER_ROOT_STRIP="/usr/local/Cellar/*/*/" # one * for the package name and one for its version
+else
+  PKG_MANAGER=port
+  # If someone knows a better way to find out where port is installed, please let me know!
+  PKG_MANAGER_ROOT="${PORTCMD%bin/port}"
+  PKG_MANAGER_ROOT_STRIP="${PORTCMD%bin/port}"
+fi
 
 ###################### UTILITY FUNCTIONS ######################
 
@@ -133,17 +140,25 @@ function add_dylibs_using_macpack {
   fi
 }
 
-# Add files from a MacPorts package using package name and grp filter
-# $1 = MacPorts package name
+# Add files from a Brew package using package name and grp filter
+# $1 = Package name
 # $2 = regexp filter (grep)
 # Note:
 # This function strips the install path of the "port" command
 
-function add_files_using_macports_package {
-  echo "Copying files from MacPorts package $1 ..."
-  for file in $(port contents "$1" | grep "$2" | sort -u)
+function add_files_of_package {
+  case $PKG_MANAGER in
+  port)
+    LIST_PKG_CONTENTS="port contents"
+  ;;
+  brew)
+    LIST_PKG_CONTENTS="brew ls -v"
+  ;;
+  esac
+  echo "Copying files from package $1 ..."
+  for file in $($LIST_PKG_CONTENTS "$1" | grep "$2" | sort -u)
   do
-    relpath="${file#${PORTDIR}}"
+    relpath="${file#${PKG_MANAGER_ROOT_STRIP}}"
     reldir="${relpath%/*}"
     mkdir -p "$RSRC_ABSDIR/$reldir"
     cp "$file" "$RSRC_ABSDIR/$reldir/"
@@ -154,7 +169,7 @@ function add_files_using_macports_package {
 # $1 = path prefix (absolute)
 # $2 = relative path to $1 and ${RSRC_ABSDIR} (must not start with /)
 
-function add_foler_recursively {
+function add_folder_recursively {
   echo "Copying files from folder $1/$2 ..."
   mkdir -p "${RSRC_ABSDIR}/$2/"
   cp -R "$1/$2/" "${RSRC_ABSDIR}/$2/"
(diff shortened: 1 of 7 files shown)


MSoegtropIMC:
@gares : is here the blocking point to test in on a Mac?

gares:
No, the problem is the the bundle does not work. My implementation of ldd is wrong, my plan was to use macpak (a python script) but I had no time to try it.

gares:
FTR the `ldd` business should be replaced by a bunch of calls calls to https://github.com/chearon/macpack

MSoegtropIMC:
As discussed, I will finalize this, but it will take a while - likely around 4 weeks.

macos/create_installer_macos.sh
```diff
@@ -0,0 +1,86 @@
...
+
+rm -rf macos_installer/
+mkdir macos_installer/
+cd macos_installer/
+
+COQ_VERSION=$(coqc --print-version | cut -d ' ' -f 1 | tr -d '\r')
+APP=_dmg/Coq_${COQ_VERSION}.app/
+
+opam source coq --dir=coq/
+eval `(cd coq ; ./configure -local >/dev/null 2>&1 ; grep VERSION4MACOS config/Makefile)`
```
MSoegtropIMC, line 13: @gares: I picked this up now. Can you explain why extracting VERSION4MACOS from the configure result is required and one can't simply use COQ_VERSION? Does this have to do with the naming of beta versions? Is the usage of COQ_VERSION above then consistent?

gares: I think apple needs x.y.z, see the comments in configure.ml

MSoegtropIMC: That's what I thought. But I think calling configure for this is a bit heavy - I will pull it from the configure.ml file instead.

MSoegtropIMC: Thinking about it: shouldn't this be the platform version?

MSoegtropIMC:
> FTR the `ldd` business should be replaced by a bunch of calls calls to https://github.com/chearon/macpack

Would you do this only for CoqIDE, so that one doesn't need the script setting LD environment variables, or would you do the patching for all executables?

MSoegtropIMC:
Regarding macpack: I experimented a bit and I will do this for all executables. The shared libs I will put into lib/dylib. The patched executables have relative paths to the shared libraries, so one can copy it to an Applications folder and it works there. Great tool!

gares:
You should run macpak on all binaries, since even coqc now depends on a shared library (libgmp), I had to fix the scripts in Coq's CI for the 8.13 release (they were only running it on coqide).

The only gotcha is that if you forget to run the script, then things may still work because you happen to have the libraries in /usr/local also in the test machine. This would have been catch by the smoke test (a la windows, in a new VM) but went unnoticed in Coq's CI + manual testing by some users.

macos/create_installer_macos.sh
gares: I'm fine with that.
The only problem is the beta, IIRC we use 8.12.91 to mean 8.13+beta1 or something like that, which is not exactly nice.

gares:
I don't recall. Globs are used only by coqdoc as far as I know.

MSoegtropIMC:
So if I don't install the standard library .glob files, coqdoc can't produce references to library symbols?

gares:
IIRC a.glob is only used to generate a.html, links to the stdlib are kind of hardwired. I don't know how it works for third libraries (not the current, not the stdlib)

MSoegtropIMC:
I see. Well let me keep them for the time being since the Windows installer also has them - compressed they are quite small. We should do some experiments on this and then decide for both.

MSoegtropIMC:
@gares : another question: does the config file mechanism you recently added also work for CoqIDE?

gares:
The environmet file? yes it does.

gares:
I may have some free time tomorrow. If you need help on this you can share your progress and tell me where I can help.

MSoegtropIMC:
OK, thanks! I did some cleanup of the folder structure and e.g. put the GTK/GDK shared libraries where they typically are, but then it doesn't work any more. I guess one has to set `GDK_PIXBUF_MODULE_FILE` as described in: (https://developer.gnome.org/gdk-pixbuf/stable/gdk-pixbuf-query-loaders.html). That is the reason why I asked about the Environment file and coqide ... Although GDK picks up a lot of stuff also from `CWD/../wherever it is in linux root` strange that it doesn't for the shared libs, but maybe there are security concerns.

In the end I would prefer if the coqide in MacOS is just a symlink rather than a shell script.

Btw.: I think you can easily use absolute paths on the Mac. In the end one puts a link to the Applications folder in the DMG and can expect people copy it there.

One open end is that I want to create an auto generated ReadeMe.Html file in the DMG root with license information. I discussed with Xavier what would be agreeable for him in terms of CompCert, and that would work for him. This also depends on (https://github.com/ocaml/opam-repository/issues/18343) unless we hack it ...

I will squash my stuff later today and push (no force push, one commit on top of yours). I am busy until mid next week, but always there for questions.

MSoegtropIMC:
@gares: I decided to push this unsquashed since some of the intermediate states might be interesting. Since I rebeased when I started, I will do a force push. If I don't hear from you I will do this 9:10.

gares:
I see, so what can I do? Maybe have CI call this?

macos/create_installer_macos.sh
```diff
@@ -0,0 +1,431 @@
...
+OPAM_FILE_WHITELIST[result]='.^'
+OPAM_FILE_WHITELIST[cppo]='.^'
+
+OPAM_FILE_WHITELIST[elpi]='.^' # linked in coq-elpi
+OPAM_FILE_WHITELIST[camlp5]='.^' # linked in elpi
+OPAM_FILE_WHITELIST[ppx_drivers]='.^' # linked in elpi
+OPAM_FILE_WHITELIST[ppxlib]='.^' # linked in elpi
+OPAM_FILE_WHITELIST[ppx_deriving]='.^' # linked in elpi
+OPAM_FILE_WHITELIST[ocaml-migrate-parsetree]='.^' # linked in elpi
+OPAM_FILE_WHITELIST[re]='.^' # linked in elpi
```
gares, line 220: I don't think we need these to have the `elpi` REPL work, nor the Coq plugin, since both link in there all dependencies.

gares:
I did `./configure -prefix=/Applications/Coq_.../Contents/Resources ;make ; make install` by hand adwanita, and CoqIDE now has icons. So I guess we are not copying the ones it wants to use on osx

gares:
I can confirm that `index.theme` is the key. If I remove it, then there are no icons. If I remove 22x22, then it still works by falling back on 16x16. I tried stripping down `index.theme` by hand, and keep only 512x512, and it picks these ones (but some are missing, maybe that size is incomplete, and is anyway too large).

I have no idea why this would be OSX specific, I checked and the index file is not in the Windows package but that one works.

MSoegtropIMC:
The selection I copy is the one from Windows - I got this by working with CoqIDE and recording all files it accesses, which I rounded up a bit. But I know that a few files are missing - I did not open the file dialog box in my tests.

I guess one should repeat such a test on MacOS - possibly with different screen resultions.

gares:
The legacy thing made icons work!

MSoegtropIMC:
> The legacy thing made icons work!

So time to test?

gares:
Well, I did also make another change which broke the scripts ;-) But yes, once it builds you can download and test if you like.

I can continue on the TODO list up there, after all I work 10 minutes every two hours on this.

The most important item is the README. You mentioned HTML: is it a feature of osx (it gets displayed automatically) or it was just an idea? I'm asking that because the snap script already generates a markdown with roughly the same piece of info https://snapcraft.io/coq-prover so I could just reuse that piece of code

MSoegtropIMC:
> You mentioned HTML: is it a feature of osx (it gets displayed automatically) or it was just an idea?

Well most people won't have anything installed which can display markdown in a nice way. if you double click on a .html file, it should open in the browser. I did explore automated markdown -> HTML converters. This works, if one keeps the markdown syntax a bit restricted (stuff like avoiding multi line verbatim inside of a bulleted list).

gares:
The last DMG is bzip2 compressed. It is not that smaller, but on this old laptop it is way slower to install.
Icons are kind of there, even if I've the feeling that the ones in the "open file" windows are not (all files look the same).

I'll look at the markdown/html tomorrow and see how it goes.

MSoegtropIMC:
Yes, the file icons are missing, also on Windows. This is because when I did the file access test, I forgot to open a file dialog. We should add this.

gares:
OK, I think the DMG is now OK, perfectible but functioning as far as I can test. I kept the TODO list up there updated.

I could not refrain for pushing here a couple of commits that could go to another PR (eg CI workflow split and snap fixes), I'm sorry but I'm not too much sensitive about VCS history cleanness.

MSoegtropIMC:
Perfect, I will try it today (after I took care of a few merges)

MSoegtropIMC:
What shall I say - it works :-)

A few observations:
- until we got it signed, we should mention in the Readme how to open an unsigned executable (Ctrl+Click on the app folder in /Applications and select open). This needs to be done only once.
- In file open / save it shows folders with a text file icon with the exception of special folders.
- tutorial_elpi_lang from the smoke test kit does not run through in CoqIDE - it fails on `Elpi Print stlc "tutorial.html"` with `tutorial.html: Read-only file system`. I am not sure where it tries to write - the test kit file is in a RW file system and I think it would make sense to use a path relative to the current file.

I am available now. Anything you want me to do? E.g. I could create a precise list of used icon files using some process monitoring tool (not 100% sure how to do this on Mac, but I guess it is not difficult).

MSoegtropIMC:
P.S.: The readme is very nice!

Btw.: I thought about a logo for Coq platform. How about the classic Coq on a typical SW box? I could create at least a preliminary version.

gares:
>     * . I am not sure where it tries to write - the test kit file is in a RW file system and I think it would make sense to use a path relative to the current file.

Hum, I think it writes in the current working directory. I guess `coqide` sets it to the app folder, while coqc does not.

gares:
I don't think I'll on this in the next days, so you can do any of the items in the todo list.
Maybe sharing code among the installers is something you can do better than anyone else.

gares:
Oh, I forgot: it may be nice to use the matrix feature in this CI as well, and install macports in one of the two. I'll update the TODO list

MSoegtropIMC:
> it may be nice to use the matrix feature in this CI as well

Yes, this is the plan, to make a matrix build with Coq 8.12, 8.13 and dev, and as you say maybe also with MacPorts and Homebrew.

gares:
I foresee a Coq 8.13.2 release (https://github.com/coq/coq/pull/14005), it would be great if we could use this DMG script.

MSoegtropIMC:
Do you have a timeline?

Right now I have some severe issues with Qed time (see the zulip discussion) and need to dig a bit into this topic, which holds me back a bit, but I could priorise it.

I think the cleanup like moving the plist file or code sharing with Windows we can do later - has no impact on the final result.

gares:
No timeline, it depends when/if that PR gets merged.

MSoegtropIMC:
@gares : As discussed, I am picking this up now.

gares:
ok, thanks

MSoegtropIMC:
@gares : a short status report: the ReadMe file looks quite nice now. although I wanted to add additional information for dependency packages, but maybe I do this in a later step.

CoqIDE works with the script shifted to bin - it is quite complicated to find the script's folder on Mac if the script is started via a symlink but it works now for all cases.

I am still thinking about improvments on the GTK side - I think there might be more general environment variables to set the root folder, e.g. GTK_EXE_PREFIX or GTK_DATA_PREFIX.

MSoegtropIMC:
P.S.: I am going to squash this into 4..6 commits. The remaining points except for the GTK environment I will do at a later point.

MSoegtropIMC:
@gares : why did you set the `XDG_DATA_HOME` variable? It doesn't seem to be required and it also looks strange that one would need this, since CoqIDE on Mac (as we build it) is not a X-Windows app.

If one puts files into the documented default places (the pixbuf loaders.cache belongs somewhere else), it seems to be sufficient to set `GTK_EXE_PREFIX`.

MSoegtropIMC:
Please ignore the above - it was bad testing (had my MacPorts folder accessible).

gares:
ok for squashing/rebasing

gares:
I'm not touching this, so you can rewrite history as you see fit

MSoegtropIMC:
@gares: I squashed essentially following our alternating work phases, but this also made nice logical blocks. I will merge in about 2 hours (if CI runs through).

I attached the latest ReadMe file - in case you want to review it.

[README.html.zip](https://github.com/coq/platform/files/6264480/README.html.zip)

PR #55, opened 2021-01-01
Smoke artifact & CI job

gares:
(includes useful commit from v8.13)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,13 +84,14 @@ jobs:
     steps:
       - name: 'Download Artifact'
         uses: actions/download-artifact@v2
+        id: download
         with:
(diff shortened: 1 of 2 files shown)


gares:
apparently osx's bash is old (v3, which does not know -A)

gares:
@MSoegtropIMC I Need help on osx, it seems I don't get to use bash in the smoke kit script. Otherwise the PR seems fine to me

MSoegtropIMC:
@gares: yes, afaik macOS doesn't include GPL3 stuff. The default is zsh. Since the script uses arrays it isn't easy to port it to plain posix, but likely not impossible either. Zsh should work, though, possibly with slightly modified syntax. I should have used Python right away ...

gares:
I fixed it by installing bash via brew on osx

PR #56, opened 2021-01-03
remove packages now present upstream

gares:


opam/packages/coq-gappa/coq-gappa.1.4.5/opam
```diff
@@ -32,8 +32,8 @@ tags: [
 authors: [ "Guillaume Melquiond <guillaume.melquiond@inria.fr>" ]
 synopsis: "A Coq tactic for discharging goals about floating-point arithmetic and round-off errors using the Gappa prover"
 url {
-  src: "git+https://gitlab.inria.fr/gappa/coq.git#3935dfe7"
-}
+  src: "https://gforge.inria.fr/frs/download.php/file/38380/gappalib-coq-1.4.5.tar.gz"
```
MSoegtropIMC, line 35: Did you discuss with @silene which of the two (INRIA gitlab vs INRIA gforge) he prefers? As far as I can tell he keeps both in sync and as far as I remember we said that the future is gitlab mostly because there were issues with the gforge path syntax and Coq's CI, although this is not an issue for opam / Coq platform. Anyway, I think we should keep this consistent.

gares: This is the URL he used in the official package, I guess he is fine with it.

silene: And that is the only URL anyway. There is no corresponding URL on Gitlab.

gares:
@MSoegtropIMC any idea about the smoke test?

gares:
I found the problem: I do not select compcert and vst when I run the installer from the command line. @MSoegtropIMC  Is there a way to do that?

gares:
I found a hint on stackoverflow, I can use silent mode to turn the sections on. I'll try that.

PR #57, opened 2021-01-07
Coq V8.13.0

gares:
Switch to Coq 8.13.0


gares:
It seems the Makefile of VST installs things in a non-standard path if ARCH is not set, I'll try to override it

gares:
I found it: patching Makefile as in https://github.com/PrincetonUniversity/VST/pull/472 is really needed since there is a call to a shell script to figure out which files need to be installer that calls back make: passing COQVERSIONS to make is not sufficient for the inner call.

PR #73, opened 2021-02-01
Cygwin64: revert MinGW 64 binutils to 2.35.1-1

MSoegtropIMC:
Revert MinGW 64 bit binutils to version 2.35.1-1.
There is some evidence that the latest version causes CI failures.

### DIFF
--- a/shell_scripts/init_cygwin_fixes.sh
+++ b/shell_scripts/init_cygwin_fixes.sh
@@ -18,7 +18,7 @@ then
   ln -s -f /usr/bin/python2 /usr/bin/python
 fi
 
-###################### Fix tar on 32 bit cygwin #####################
+###################### Fix tar in 32 bit cygwin #####################
 
 # Cygwin updated to tar 1.33 which is broken on Cygwin 32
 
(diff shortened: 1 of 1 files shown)


shell_scripts/init_cygwin_fixes.sh
```diff
@@ -29,3 +29,15 @@ then
...
 fi
+
+###################### Fix binutils in 64 bit cygwin #####################
+
+# Cygwin updated to mingw64-x86_64-binutils is broken on Cygwin 64
+
+if [[ "$OSTYPE" == cygwin ]]
+then
+  if [ "`uname -m`" = "x86_64" ]; then
+    wget http://mirrors.kernel.org/sourceware/cygwin/x86_64/release/mingw64-x86_64-binutils/mingw64-x86_64-binutils-2.35.1-1.tar.xz -O /tmp/mingw64-x86_64-binutils-2.35.1-1.tar.xz
```
gares, line 40: As per zulip, I think you should be resilient to the file not being there.

MSoegtropIMC: As is the script will stop with an error if the file is not there. I think this is what we want. The main script has `set -o errexit` (in a sourced init script).

PR #75, opened 2021-02-02
Cygwin32: remove tar patch - seems to be fixed upstream

MSoegtropIMC:
There is a new release of tar in Cygwin, which likely fixes the observed issues for Cygwin 32.

Update: looks like this was a misunderstanding.


MSoegtropIMC:
The tar package is not yet fixed upstream in Cygwin, so I will close this for now.

PR #81, opened 2021-02-22
Release 8.13.1

gares:


gares:
@MSoegtropIMC I believe it may be also the time for releasing/tagging 8.13.1.0 (I would switch to the date based versions from 8.14)

MSoegtropIMC:
@gares : yes. What held me back is watching the failures of the master branch nightly CI. Some of these would also affect release branches, since they are due to changes in opam packages and/or cygwin. So I wonder if we shouldn't fix all opam package versions (not pin them, but to request a fixed version) and also maybe to point to our own cygwin repo.

MSoegtropIMC:
P.S.: since there was no non beta release of Coq Platform 8.13 so far, I planned to use the new format.

gares:
Hum, I'm OK with that too, but I then need to ask for a manual update in the snap store.

gares:
I see only a coq-interval problem, fixed by this release of Coq, and pkg-config one which I don't know really know what to do about.
Which are the packages that need an update?

MSoegtropIMC:
@gares : there is not need to update packages, but it would make it more stable to install a well known version of all packages. It does happen that new versions of opam packages break things, e.g. conf-pkg-config.

MSoegtropIMC:
For pkg-config I have a fix in master - you just have to copy over the local patch package.

gares:
let me do that

Zimmi48:
> Hum, I'm OK with that too, but I then need to ask for a manual update in the snap store.

Related to this: it would be good to get out of beta for the Snap package. The creation of the new track seems to me like a good time to do this.

gares:
@MSoegtropIMC  I've seen this already but I can't recall what is wrong.
```
====================== Running test file coq-vst/verif_reverse2.v ======================
coqc  verif_reverse2.v
File "./verif_reverse2.v", line 6, characters 15-34:
Error: Cannot find a physical path bound to logical path matching suffix
VST.floyd.
```

MSoegtropIMC:
> Running test file coq-vst/verif_reverse2.v

I need to check this. This used to be an error in picking the wrong 32/64 bit variant of the file - there are two files with the same name in VST from a progs and a progs64 folder and the smoke test kit has to pick the correct one - but it looks correct in the v8.13 branch. I will do a test build and have a look.

MSoegtropIMC:
@gares : regarding "Running test file coq-vst/verif_reverse2.v" : the issue is that the VST make install does not install anything at all. This used to work a few weeks back - need to check.

MSoegtropIMC:
@gares : as it looks, the "IGNORECOQVERSION" messes up make install for the 64 bit version. I am currently doing a local test build with the traditional method of fixing this (which is to patch the VST makefile to allow the new Coq version). If this works I will change it in the opam repo directly and remove the local patches in your PR branch.

I will do a PR for VST to check in the future only the major version of Coq - not sure, though, what would be a good platform independent way of doing this in make. I guess one could remove from a string postfixes of the form ".0, .1, .2, +beta1, +beta1" and compare the rest. If you aware of a good example, please let me know.

gares:
My take on that is that it is not the duty of the Makefile to test for the right Coq version, for two reasons:
- that metadata is in opam already, and opam has a better language to express a range, and options to force/ignore a version constraint
- users should `opam install .` and not `make && make install` (and in general `opam install coq-foo`)
At best, the makefile could test if it is being run by opam or not and, if not, print a warning "unknown/untested version of Coq".

I know we had very bad records about retro compatibility, I cannot blame compcert or VST for being cautios. Imo the IGNORECOQVERSION thing is OK (modulo bugfix)

MSoegtropIMC:
I think the issue is that Andrew has a few users which use the makefile directly for modified versions. The makefile has quite a few more options than the opam package. For this reason the sanity checks in the VST makefile are expected to be reasonably self contained.

gares:
My problem is that I've uploaded the windows installer in the release of Coq. I've added a note that VST is not in there. Are you working on a fix right now? If not I'll do it (for this branch I mean)

MSoegtropIMC:
I am currently investigating the problem. Using the original opam file with the makefile patched (as done for 8.13.0) does not work either. I currently don't understand what is going on here. I have a 8.13 opam switch from February 16th where VST does work. The opam file did not change in between.

This might take a few hours since the build time for VST is quite long.

gares:
OK, great! (I thought the diagnostic was clear already)

MSoegtropIMC:
I did a silly mistake - I forgot to `opam update` the local patch repo after changing the opam files for VST - a common error when switching between MacPorts and opam work on the same day, since MacPorts scans local patch repos automatically. So the fix I proposed seems to work.

Is this super urgent? If so I can push it right away. Otherwise I would prefer to make a clean local build first.

gares:
not super urgent, if you go it today its fine, otherwise I'll do it.

MSoegtropIMC:
Should be within two hours.

As I said I will change the opam files in the opam coq repo and then remove the local opam patch in your branch.

MSoegtropIMC:
@gares : FYI: local test are successful. I created (https://github.com/coq/opam-coq-archive/pull/1634). I expect that opam CI takes on hour or so.

gares:
great!

MSoegtropIMC:
@gares : is this ready to merge from your side? CI went through now.

I find it a bit strange, that you did put the Coq packages here as local packages. Why not put them upstream right away? I guess this counts as non controversial and can be self merged immediately by the release manager. But if your prerfer it this way, we can do it this way as well.

MSoegtropIMC (approved):
Except that I would put the Coq packages upstream right away (no very strong feelings about this) this PR is fine.

gares:
they are upstreamed already!

gares:
I will remove them

MSoegtropIMC:
Perfect. I guess it is time to formulate some release message for Coq Club then :-)

MSoegtropIMC:
One thing I forgot: we need to update the ReadMe.md file since it contains the actual package versions. I will do so and then merge without rerunning CI.

PR #84, opened 2021-02-24
Version 2021.02.0

MSoegtropIMC:
- Bump version to 2021.02.0
- Cleanup of README files


Zimmi48:
My impression is that part of the Snap package documentation is oriented toward developers, so it might not be fully appropriate as a user documentation in the Linux README.

MSoegtropIMC:
@Zimmi48 : I found the documentation quite readable and user oriented and we do advertise binary installers for Linux.

@gares: what is your opinion?

I merge now so that we can judge the final product, including links, ...

gares:
I think there are a few sentences which are relevant for developers only, but maybe it is not enough to justify another readme. So I'm ok with this pr

PR #86, opened 2021-02-25
Add a sort of contents table to the Linux README.

Zimmi48:


MSoegtropIMC:
Should we have the same lines of introduction for Windows?

Zimmi48:
It already contains the following sentence and the first section is short, so it looks good to me as is:

>This method is intended for beginners. For advanced users it is recommended to compile Coq from sources via opam (see next section).

PR #88, opened 2021-02-25
fix smoke test for HB

gares:


MSoegtropIMC:
I already had a PR for it, but since you added one more file I will cancel mine and we take yours.

gares:
Sorry, I still have to wake up fully ;-)

gares:
My bet is that `readme.v` requires no "sed line" in order to work if HB is installed system wide, let's see if CI agrees.

MSoegtropIMC:
Well you anyway know best which files to pick - I more or less throw dice when picking smoke test files.

I thought about having something in opam to denote a "look at this first example file". I think it would in general be useful information for users.

gares:
I think you did a good job and I did advertise your script to a jscoq dev wanting to do a smoke test as well.

W.r.t. HB you picked the hardest example, which is ok. The one I've added, `readme.v`, is the one we use for the `README.md`: it is much simpler but we try to keep it up to date.

MSoegtropIMC:
@gares : I think we need something like:
```
TEST_FILES[coq-hierarchy-builder]='demo2/classical.v demo2/stage10.v demo2/stage11.v'
COQ_OPTION[coq-hierarchy-builder]='-Q . demo2'
```

gares:
Yes, but I also adjusted the regexp.

MSoegtropIMC:
Ah ok, I forgot about teh AWK script. I think one or the other should be good enough (either coq options or patching the requires).

gares:
finally ok

PR #97, opened 2021-03-29
Customize scripts for a custom coq-hott-platform

mattam82:
Let's try the CI here, I didn't manage to make it run on my fork.


coq_platform_make.sh
```diff
@@ -42,10 +42,16 @@ source shell_scripts/sanitize_environment.sh
...
+COQ_PLATFORM_EXTENT=f
+# source shell_scripts/ask_introduction.sh
+COQ_PLATFORM_PARALLEL=p
+COQ_PLATFORM_JOBS=4
+#source shell_scripts/ask_parallel_build.sh
+COQ_PLATFORM_COMPCERT=n
+#source shell_scripts/ask_compcert_open_source.sh
+COQ_PLATFORM_VST=n
+# source shell_scripts/ask_vst.sh
+COQ_PLATFORM_SWITCH=d
```
gares, line 54: FYI: you can pass these to the platform script, see PLATFORM_ARGS in .github/workflow/*.yaml

mattam82: Yep, I realized later :)

coq_platform_packages.sh
```diff
@@ -27,39 +27,41 @@ if  [[ "${COQ_PLATFORM_EXTENT}"  =~ ^[fF] ]]
...
+#PACKAGES="${PACKAGES} coq-mathcomp-character.1.12.0"
 # Plus a few extra mathcomp modules
-PACKAGES="${PACKAGES} coq-mathcomp-bigenough.1.0.0"
-PACKAGES="${PACKAGES} coq-mathcomp-finmap.1.5.1"
-PACKAGES="${PACKAGES} coq-mathcomp-real-closed.1.1.2"
+#PACKAGES="${PACKAGES} coq-mathcomp-bigenough.1.0.0"
+#PACKAGES="${PACKAGES} coq-mathcomp-finmap.1.5.1"
+#PACKAGES="${PACKAGES} coq-mathcomp-real-closed.1.1.2"
+
+PACKAGES="${PACKAGES} coq-hott.dev"
```
gares, line 61: According to https://github.com/coq/opam-coq-archive/blob/master/extra-dev/packages/coq-hott/coq-hott.dev/opam you need `= "coq.dev"` while here you have `coq.8.13.1`. You can either pick a version of `coq-hott` which works on 8.13, or put your own "overlay coq-hott package" in `./opam/packages/` and tune the deps there.

mattam82: Right!

mattam82: I added a custom made 8.13 and updated the dependency. let's see!

gares:
It is better but the smoke test fails, you have to disable it and its upload I m afraid.

mattam82:
> It is better but the smoke test fails, you have to disable it and its upload I m afraid.

Yep, I'm trying to just disable it for coq-hott (it would require running `hoqc` rather than `coqc`)

mattam82:
On top of PR #50 to get dmg packages as well.

gares:
> On top of PR #50 to get dmg packages as well.

FYI the split of CI in separate workflows did duplicate COQ_PLATFORM_ARGS, there is one instance per OS, so you may want to strip also osx and not just windows

gares:
Also, the snap package does only build a "slim" version of the bundle, which is larger than the other ones, and does not save the artifact. You have to manually run a workflow from the actions menu with appropriate options if you want to get one. It is a bit unclear to me how you could distribute it, since the snap store is not really for this kind of variants (unless they become official in some sense). You can still upload the snap somewhere and have people download install a .snap file from the command line, that works.

mattam82:
Yep I was thinking of just distributing all the packages from somewhere on the web.

MSoegtropIMC:
So for Windows this works. For Mac one has to remove the patching of the HoTT config file, since it doesn't exist any more. I can take care of this. Can someone look into the Snap package?

Btw.: HoTT should now be a completely normal opam package, so maybe it would be easier to start from scratch based on @gares DMG PR.

MSoegtropIMC:
@mattam82 : since HoTT is now a simple opam package, I will close this PR. I will add HoTT to Coq Platform 2021.02.1.

Issue #92, opened 2021-03-17
opam-repository-mingw will be discontinued as of Aug 2021


Zimmi48:
https://fdopen.github.io/opam-repository-mingw/2021/02/26/repo-discontinued/
Are you aware of this?
How does it affect us?

MSoegtropIMC:
Ah thanks - I didn't see this!

How is the state of INRIA's effort for better OCaml support on Windows?

It wouldn't be much work to fix all packages we need so that they work on Windows on the main opam repo, but I am still using a binary ocaml and opam delivery from that repo. But then I did build this from sources before and could do so again.

We don't depend on depext for cygwin - instead we install all required packages upfront.

I see that this is finished before the repo is out of maintenance.

Zimmi48:
>How is the state of INRIA's effort for better OCaml support on Windows?

This effort was initiated by @yurug, who is not at Inria anymore (he's moved to Nomadic Labs) but he's probably still a member of the board of the OCaml Foundation.

But now, as far as I know, this effort is mostly led by OCaml Labs. Maybe @avsm can inform us of the progress and plans.

yurug:
Yes, I agree with Théo: @avsm is currently the most knowledgable about this
important project.
-- 
Yann Régis-Gianas

MSoegtropIMC:
So how shall we proceed? Wait or just get it done? What needs to be done is:

- fix the opam OCaml packages, so that they work on Windows (cygwin MinGW cross)
- fix all other opam packages we need in the main opam repo - from an opam point of view there is no good reason to keep a separate repo. Opam has sufficient logic to make packages work on all platforms.
- Have some way to create a MinGW on Cygwin cross opam, preferably make this a Cygwin package

Did I forget anything?

MSoegtropIMC:
Maybe we should do a test of all opam packages and for those which don't compile on Windows set corresponding availability information and then fix those we need.

Zimmi48:
An update: https://discuss.ocaml.org/t/ann-ocaml-opam-images-for-docker-for-windows/8179
And a related discussion: https://discuss.ocaml.org/t/ocaml-for-windows-installation-confusion/8105

I should have cc-ed @MisterDA since he has kept working on Windows support for OCaml (now at Tarides).

MisterDA:
Opam 2.1, soon to be released, will have improved support for Windows. We expect to be able to migrate completely to opam 2.1 and opam-repository, maybe with a few rough edges at the beginning, but it should work.
[...]

PR #102, opened 2021-04-18
Multi version

MSoegtropIMC:
This PR enables support for selecting a package list version. Currently supported are:
- 8.13
- 8.12
- dev

Reenabling the dev branch seems to have issues on Mac, so let's see what CI says ...

@gares : except for the issues with the dev branch, this is very close to final.


Zimmi48:
If switch names support `+`, I would really recommend using it before the Coq version to distinguish it from the platform version.

MSoegtropIMC:
> If switch names support +, I would really recommend using it before the Coq version to distinguish it from the platform version.

Yes, I thought that the . gets hard to read. I will try "+". I also thought about using "_", but I like "+".

Zimmi48:
One issue with `+` than I didn't realize when commenting is that if we keep using it for beta versions there might be switches with two `+` signs that do not mean the same thing (`2021.07+beta1+8.13`). So we should either stop using it for beta versions (`2021.07beta1+8.13`) or use a different symbol for the Coq version (`2021.07+beta1_8.13`).

MSoegtropIMC:
So how about "__coq-platform_2021.02.1_8.13" so using "_" also es separator between coq-platform and its version? Btw. It used to be "_coq-platform_.2021.02.1". I decided to start with two underscores as classical prefix for "system reserved names in a user name space" but don't have strong opinions about this.

Zimmi48:
Yes, your proposal looks pretty great!

gares:
Overall it looks ok, but since you have coq_packages.8.12.sh I guess I don't get exactly what the plan is. 2021.02.x was, in my mental model, pivoted on Coq 8.13.x and, at least for some packages, there is no choice (coq-elpi 1.9 does not work on 8.12, you need 1.8 for that). Maybe you want to have coq_package_2021.02.x.sh  and, say, coq_packages_2021.06.x.sh in the same repo, each pivoted on its coq version, but this PR fixed one platform version and multiple package lists.

Can you clarify this and how git branches would then be used?

MSoegtropIMC:
@gares: For packages which did exist for older versions of Coq the idea is to add them to the respective package list - for packages which are new for Coq 8.X they are only added from 8.X onward. So the idea is that the package lists do increase also for old versions.

After pondering over this for a while I think that this is the easiest to maintain and most useful method for users.

I am not yet sure what the final branch structure will be. Most likely this will become the master branch and we release from the master branch since with a multi-version branch there is no good reason for doing releases from older Platform versions.

gares:
I'm not so sure. What sould 2021.02.1 mean if by picking 8.12 or 8.13 the versions of the packages change?

MSoegtropIMC:
> I'm not so sure I get it. What would 2021.02.1 mean if by picking 8.12 or 8.13 the versions of the packages change?

Well yes, one has to say `2021.02.1_8.12` or `2021.02.1_8.13` or `2021.02.1_dev` and probably would use `2021.02.1` as abbreviation for `2021.02.1_8.13`.

gares:
Maybe It's just a detail, but I'd prefer `2021.02.1` (for 8.13) and `2021.02.1_8.12` for the "backport" thing (`2021.02.1~8.12` if made an opam package, so that it is `<` than the one without a coq version). If only because we can't change the version schema every day ;-)

MSoegtropIMC:
It is an important detail and I agree that we should fix the naming now. IMO it is perfectly agreeable to say "The Coq Platform" is the one with the latest released Coq and then there are "compatibility versions" with newer (dev - maybe some day even a beta) or older Coq. So shall we go with that?

And indeed it does make sense to come up with a naming scheme, which would work as a name for opam packages as well - I didn't think about it.

I guess one would have a dev Coq Platform package in `extra-dev` but everything else should work naturally. So I guess if we use the short name for the latest release, the previous versions must be named e.g. "~8.12"? How about the dev version?

Zimmi48:
>IMO it is perfectly agreeable to say "The Coq Platform" is the one with the latest released Coq and then there are "compatibility versions" with newer (dev - maybe some day even a beta) or older Coq. So shall we go with that?

:+1: 

>And indeed it does make sense to come up with a naming scheme, which would work as a name for opam packages as well - I didn't think about it.

I'm actually wondering if we will ever want to provide a meta opam package for the Coq platform. The scripts as they are, with their management of opam switches, are pretty convincing. In particular, the Coq platform pins the version of the underlying OCaml compiler. opam packages for the Coq platform would probably be worth the trouble only if we were to relax this constraint, right?

MSoegtropIMC:
> I'm actually wondering if we will ever want to provide a meta opam package for the Coq platform. 

There are currently a few issues with this:
- I have patch opam packages required for Windows which I couldn't get accepted upstream. Likely the best fix for this would be to create a fixed gtk-sourceview package for cygwin. There might be other ways around this after introducing the local pkg-config concept in opam.
- I have .dev packages for OCaml stuff (menhir, gappa) and main opam doesn't have an extra-dev concept. I guess here the best solution would be to add them to Coq's extra-dev.
- The system sanity checks are helpful - on all systems. One would have to get some of this into opam or create a "check my system sanity" opam package.

Also I think it would help to get the Windows opam repo merged into the main opam repo - as is the patch repo is required from time to time, since without it merge delays are not manageable. And I don't think it would be good to change recommendations every other release.

So this is a goal, but not short term.

MSoegtropIMC:
So the final conclusion is to name the switches:

- `__coq-platform.2021.02.1`
- `__coq-platform.2021.02.1~8.12`
- `__coq-platform.2021.02.1~dev`

provided this works. Note that I went back to have a dot between `coq-platform` and the version, because this is what one would use with `opam install` as well. So the package name would be the same except that the two underscores are removed. I would like to keep this, because as I said this are auto generated names in a user name space.

Zimmi48:
>Also I think it would help to get the Windows opam repo merged into the main opam repo - as is the patch repo is required from time to time, since without it merge delays are not manageable. And I don't think it would be good to change recommendations every other release.

About this, have you noted #92?

MSoegtropIMC:
@gares : I am looking at the snap file and wondering why it doesn't use the full coq platforms script. I guess it is required to install opam on its own, e.g. cause it needs to live in a specific folder, but one could still setup opam upfront and let the script setup the switch and the repos and do all the rest. Is there a good reason why you didn't do that?

gares:
I think it does for the build process:
https://github.com/coq/platform/blob/2021.02/linux/snap/snapcraft.yaml.in#L23

It uses bits and pieces of it to generate the package description, IMO. I did that before realizing it is not updated automatically.
Is this what you are talking about?

MSoegtropIMC:
@gares : thanks - O just didn't understand it. I will fix it then. I guess one can in a later PR change it such that it uses the same packaging logic as Windows and Mac, so that it would also work with manually extended opam switches.

gares:
The best would be to:
- minimal sed on snapcraft.yml.in with no need to have opam & co around, just set the version string and the platform script flags
- from the build step of the snap
  - run the platform script
  - run a snap specific installation script (now inlined and not shared with win/mac, e.g. the blacklist of files etc)
  - print the package description for the publishing step (similar code to the one generating html for mac)
  
Just in case, we can't run the main platform script outside snap, and then copy the files inside, because it is snap which sets up gtk & co, and it has to stay that way, since installing the snap pulls in the gnome "platform" snap as a dependency.

The snap really contains an opam root, properly trimmed. We can take the files from there and putting them in a mirror layout, as we do in the other install scripts, if it comes handy to share more code with win/mac it is welcome, but if not there is really no need for it.

gares:
let me merge some commits in hierarchy-builder/coq-master

opam/opam-coq-archive/extra-dev/packages/coq-hierarchy-builder/coq-hierarchy-builder.dev/opam
```diff
@@ -0,0 +1,22 @@
...
+depends: [ "coq-elpi" { (>= "1.10.2" & < "1.11~") | = "dev" } ]
+conflicts: [ "coq-hierarchy-builder-shim" ]
+synopsis: "High level commands to declare and evolve a hierarchy based on packed classes"
+description: """
+Hierarchy Builder is a high level language to build hierarchies of algebraic structures and make these
+hierarchies evolve without breaking user code. The key concepts are the ones of factory, builder
+and abbreviation that let the hierarchy developer describe an actual interface for their library.
+Behind that interface the developer can provide appropriate code to ensure retro compatibility.
+"""
+tags: [ "logpath:HB" ]
```
gares, line 22:
```suggestion
tags: [ "logpath:HB" ]
url {
      src: "git+https://github.com/math-comp/hierarchy-builder#coq-master"
}
```

gares: This is the problem: if there is no URL stanza, then it probably fetches some old version which does not have the `build` make target (as the log report)

gares: CC @MSoegtropIMC

MSoegtropIMC:
> ~let me merge some commits in hierarchy-builder/coq-master~ No, this is not the problem. I believe the problem is the overlay which is incomplete.

Yes, I copied the opam file from hierarchy-builder master without much thinking. Thanks for the fix!

MSoegtropIMC:
@gares : as discussed, I won't merge this into 2021.02 but instead push this as master and make this the new main branch then.

gares:
OK, works for me

gares:
maybe call this branch `main` (the new default on github)

MSoegtropIMC:
Thanks for the hint - I will call it `main`!

gares:
The snap is still disabled. What was the problem there?

MSoegtropIMC:
It is not compatible with the multi version stuff. I fixed a few things (e.g. opam repo layout) but not all. E.g. one needs to be able to recover the name of the version file from the switch name and the like. So the names / interfaces of some scripts the snap packager uses have changed. It shouldn't be hard to fix (for me since I know what changed), but I didn't have time over the weekend.

MSoegtropIMC:
This went into the main branch and won't be merged.

PR #103, opened 2021-04-30
argument -override-dev-pkg=

gares:
This PR adds a command line argument to the platform script to override on the fly the URL from which the sources of a .dev package are downloaded. This (or something equivalent) is needed in Coq CI since otherwise the platform job does not use the sources of Coq being part of the PR (but master).

CC @MSoegtropIMC , see also https://github.com/coq/coq/pull/14217


shell_scripts/coq_platform_override_dev_pkg.sh
```diff
@@ -0,0 +1,24 @@
...
+
+###################### OVERRIDES #####################
+
+for override in ${COQ_PLATFORM_OVERRIDE_DEV:-}
+do
+    pkg="${override%%=*}"
+    url="${override#*=}"
+    echo "Overriding $pkg.dev url with $url"
+    mkdir -p opam/packages/$pkg/$pkg.dev/
+    opam show --raw $pkg.dev | tr -s '\n' ' ' | sed "s@ src: [^ ]*@ src: \"$url\"@" > opam/packages/$pkg/$pkg.dev/opam
```
gares, line 19: I think we should also replace any 2 or more spaces with just one, and make the sed line also accept no space after `:`. I could not find an easier way to "edit" an opam file field.

MSoegtropIMC: I would use `sed "s@^ *src: *[^ ]*@ src: \"$url\"@"`

MSoegtropIMC (approved):
LGTM - enhancements to the sed pattern are minor since we control the opam files in question. A safer solution which also includes the enclosing braces is hard to do in a portable way with sed (awk might work better).

Please handle the sed pattern as you see fit and merge.

MSoegtropIMC:
P.S.: I just saw that coq-dev also fails (on first glance I thought that only extended packages fail). I must admit I am not sure what the state of the dev-ci branch is - I am working on enabling dev builds in the 2021.02 branch.

gares:
ok, then I'm merging this to unblock coq/coq#14217

PR #104, opened 2021-05-11
Rename CoqIde to CoqIDE and add icon

JasonGross:
Also remove coqtop from the start menu, I think it'll just confuse users, especially as it's installed as "Coq"

See https://coq.zulipchat.com/#narrow/stream/250632-Coq-Platform.20devs.20.26.20users/topic/Windows.20icons/near/238308109

Is this the right branch to propose for?

### DIFF
--- a/windows/Coq.nsi
+++ b/windows/Coq.nsi
@@ -100,8 +100,8 @@ Section "-CoqWinStuff" Sec_CoqWinStuff
   CreateDirectory "$SMPROGRAMS\Coq"
   ; The first shortcut set here is treated as main application by Windows 7/8.
   ; Use CoqIDE as main application
-  CreateShortCut "$SMPROGRAMS\Coq\CoqIde.lnk" "$INSTDIR\bin\coqide.exe"
-  CreateShortCut "$SMPROGRAMS\Coq\Coq.lnk" "$INSTDIR\bin\coqtop.exe"
+  CreateShortCut "$SMPROGRAMS\Coq\CoqIDE.lnk" "$INSTDIR\bin\coqide.exe" "" "$INSTDIR\bin\coqide.exe" 0
+  ; CreateShortCut "$SMPROGRAMS\Coq\Coq.lnk" "$INSTDIR\bin\coqtop.exe"
   WriteINIStr "$SMPROGRAMS\Coq\The Coq HomePage.url" "InternetShortcut" "URL" "http://coq.inria.fr"
   WriteINIStr "$SMPROGRAMS\Coq\The Coq Standard Library.url" "InternetShortcut" "URL" "http://coq.inria.fr/library"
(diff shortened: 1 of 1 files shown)


gares:
@MSoegtropIMC is reorganizing branches, so I let him decide where to merge.
@JasonGross could you test the artifact, eg https://github.com/coq/platform/suites/2714267362/artifacts/59864169 (I don't have a windows machine at hand)

JasonGross:
This is very strange, the icon shows up in Windows Explorer but not the start menu.... perhaps there's some windows cache that needs to be updated?
image


In any case, I think this is already an improvement.

Hm, but the program titlebar shows `CoqIde`, not `CoqIDE`?
image

JasonGross:
Hm, this does not seem to work well enough (even though it is an improvement).  I see
image
If I open file location I do see the icon, and when I click properties I see
image
change icon gives
image
and if I click "ok" the little shortcut arrow on the icon disappears:
image
clicking ok again does not change the icon in windows explorer:
image
but somehow it does change it in the start menu!?
image

gares:
Maybe try to uninstall and reinstall

JasonGross:
Tried that twice, behavior is reproducible

JasonGross:
Hm, but refreshing windows explorer with F5 seems to fix it?  My guess is that the issue in https://github.com/coq/platform/pull/104#issuecomment-839945355 is that the icon file only has one size of icon, and it's not the right size for the large start menu?

JasonGross:
In fact coqide.exe seems to contain only a 42x64, 32-bit icon, whereas other apps I have contain, e.g., all of 256x256, 128x128, 64x64, 48x48, 32x32, 24x24, and 16x16

MSoegtropIMC:
Indeed the Icon could be improved (also in other ways ;-). I somewhere have an SVG version of it.

gares:
This png probably already ok https://github.com/coq/coq/blob/master/ide/coqide/coq.png 
Anyway, I think the culprit is https://github.com/coq/coq/blob/master/ide/coqide/coq_icon.rc which points to a single .ico file

JasonGross:
Single ico files can store multiple sizes of icons, and indeed can store multiple unrelated icons (hence why we use an index of 0 here).

JasonGross:
I can prepare a PR to Coq for adding multiple icon sizes

JasonGross:
Or according to https://legacy.imagemagick.org/Usage/thumbnails/#favicon you should be able to do
```
convert coq.png -alpha off -resize 256x256 \
          -define icon:auto-resize="256,128,96,64,48,32,16" \
          coq.ico
```

JasonGross:
We maybe want `-background transparent` instead of `-alpha off|

gares:
Merged, thanks. I won't make a 8.13 point release just for this. IMO we can patch coqide at the opam level here by using files: https://opam.ocaml.org/doc/Manual.html#files

gares:
I forgot about this PR, I guess it is still relevant (even if the icon update was merged upstream), right?

MSoegtropIMC (approved):
LGTM, will merge now.

Issue #116, opened 2021-08-02
2021.02.1 doesn't install with most recent version of opam


k4rtik:
```
===== INSTALL OPAM DEPEXT =====
2.1.0
[ERROR] depext unmet availability conditions, e.g. 'opam-version >= "2.0.0~beta5" & opam-version < "2.1"'
```

Zimmi48:
FTR this is affecting the CI builds for Ubuntu and Snap.

PR #118, opened 2021-08-09
Add UniMath to Coq Platform

maggesi:


coq_platform_packages.sh
```diff
@@ -39,6 +39,9 @@ PACKAGES="${PACKAGES} coq-quickchick.1.5.0"
 # Homotopy Type Theory (HoTT)
 PACKAGES="${PACKAGES} coq-hott.8.13"
 
+# Univalent Mathematics (UniMath)
+PACKAGES="${PACKAGES} coq-unimath.20210805"
```
maggesi, line 43: Fix version.
```suggestion
PACKAGES="${PACKAGES} coq-unimath.20210807"
```

Zimmi48:
FWIW, the Snap and Ubuntu failures are expected (#116) but the Windows failure is related to this PR. It is a:

```
Fatal error: out of memory.
```

while building UniMath.

benediktahrens:
@Zimmi48 : thanks a lot for your feedback.

Is there a way to see which specific thing in `UniMath/Bicategories/DisplayedBicats/FiberBicategory/FiberBicategory1.v` causes the memory error?

Is there a way to know whether the same build error will occur on Ubuntu or Snap? What situations require a 32-bit version of Windows?

Reducing the memory usage in the affected files to an acceptable (for the purpose of the present build) will be difficult, due to the cumbersome way of testing. (I don’t have a Win machine myself, and would need to test by filing PRs here.)

I see the following workarounds:
- excluding the offending file
- excluding the UniMath package “Bicategories”
All of these seem to require publishing separate OPAM packages. 
Are there any other solutions or workarounds that I am overlooking?

maggesi:
Reducing the memory usage can be a hard work.
I propose the following compromise: include a reduced version of UniMath on Win32.

I'm not sure which is the best way to do that.
Perhaps, I can prepare a second opam package "unimath-small" and instruct the build script to conditionally include it instead of the regular one on win32.

@Zimmi48: Are there alternative/better solutions?

Zimmi48:
Yes, this seems like a good strategy to me (sorry for not responding earlier). @MSoegtropIMC do you agree?

MSoegtropIMC:
I would have to test it before I can comment on this, but basically yes.

I guess UniMath requires a special standard library similar to HoTT?

benediktahrens:
> I guess UniMath requires a special standard library similar to HoTT?

Hi there! UniMath avoids loading the Coq prelude by using the option `-noinit`. I am not sure if that is similar to what HoTT does.

benediktahrens:
@MSoegtropIMC : could you provide some guidance as to how to implement such a conditional switch? Are there any other packages that use such a switch, and that we could look at?

MSoegtropIMC:
HoTT had traditionally shell script wrappers for everything. As far as I remember some minor additions to the capabilities to _CoqProject files have been made, so that one can give the `-noinit` in a CoqProject file, so that CoqIDE and other UIs work without issues.

MSoegtropIMC:
Can you please rebase so that CI is rerun - it should work then. I won't merge it to 2021.02, though, but we will do a 2021.09 soon.

maggesi:
@MSoegtropIMC : I did the rebase.
I'm available to contribute to solve the issue on Win32.
What I can try to do is to provide a mechanism of conditional compilation depending on the architecture.
Let me know if you have suggestions about this.

MSoegtropIMC:
@maggesi : The issue on 32 bit Windows is an out of memory and might not be easy to solve - I guess you are going beyond the 2GB process limit on 32 Bit Windows.

We don't have this as yet, but we can discuss to have certain packages only on 64 bit Windows. We would have to check that it is stable on Windows 64 CI, though.

Of course if you have ideas on reducing memory consumption ...

Zimmi48:
> We don't have this as yet, but we can discuss to have certain packages only on 64 bit Windows. We would have to check that it is stable on Windows 64 CI, though.

This, or install a unimath-small on 32 bit Windows as was proposed above.

benediktahrens:
I would prefer not supporting Win32 over having a separate version for it. The 32bit version seems to be on its way out (https://www.engadget.com/windows-10-32-bit-oem-173055990.html), and having separate versions could cause confusion.

MSoegtropIMC:
There are some Coq users which prefer the 32 bit version, because it is faster (Coq is almost all pointers and these are half the size = half the cache on 32 bit), but I fully agree that the days of 32 bit are coming to an end.
So I will add a mechanism to disable certain packages on certain platforms.

Zimmi48:
@MSoegtropIMC It seems to me that you've forgotten about UniMath because there was no issue tracking it. But here, the package inclusion request comes directly from the maintainers.

MSoegtropIMC:
@Zimmi48 : well spotted, thanks!

Regarding 32 bit Windows issues: I would summarize the conclusion such that we won't have it on 32 bit Windows - which I believe is an acceptable rule to our cross platform guarantees.

MSoegtropIMC:
@maggesi : since we are about to change the default branch and had a lot of changes meanwhile, I will cherry pick your commit into one of my PRs and discard this one.

benediktahrens:
@Zimmi48 @MSoegtropIMC : Thank you very much for your work on the Coq Platform!

MSoegtropIMC:
@maggesi @benediktahrens : a question from a helicopter view follower of univalent mathematics and HoTT: as far as I can tell there is a large overlap between the two fields. Can you point me to a (beginner level) article describing the differences?

MSoegtropIMC:
In (https://github.com/coq/platform/pull/147) I cherry picked your commit and also added UniMath to the 8.14+beta1 (it compiles fine there and the smoke test you provided works).

Thanks for providing the PR - it helped a lot especially the smoke test kit file and parameters!

I will close this PR.

benediktahrens:
> @maggesi @benediktahrens : a question from a helicopter view follower of univalent mathematics and HoTT: as far as I can tell there is a large overlap between the two fields. Can you point me to a (beginner level) article describing the differences?

I am not aware of an established definition of the two things, in particular, anything written up on the subject.(*) I can briefly sketch how I use the terms.

a) HoTT, for me, refers mostly to (i) the field of research studying homotopy-theoretic (in terms of weak factorization systems, for instance) semantics of type theory, and to formalizing homotopy-theoretic results *synthetically* in type theory.

b) Univalent foundations and univalent mathematics describes, for me, the idea of using univalent type theory as a foundation for all of mathematics.

Regarding the different libraries of computer-checked proofs:

The UniMath library follows the idea of b): there is all sorts of mathematics formalized there, in particular, ordinary set-theoretic mathematics. There is no, or little, synthetic homotopy theory in UniMath, since that typically relies on Higher Inductive Types.

I don't know what exactly is available in the HoTT library that is already part of the Coq Platform. Certainly the contents are not restricted to synthetic homotopy theory.

(*) I would be very grateful for any pointers myself.

MSoegtropIMC:
@benediktahrens : thanks for the answer! Trying to summarize it in my own words: both are using a very similar or even identical foundation, but are constructing quite different buildings with even different purposes on top of it: general mathematics vs. type theory. So the lower floors - the "interface to the foundation" - are quite similar, but the upper floors are quite different. Is this a reasonable image?

benediktahrens:
@MSoegtropIMC : I would say "general mathematics vs. synthetic homotopy theory". 
I think this image you describe is reasonable, but I want to emphasize that this is only my own view. Others might use the terms HoTT and UF/UM differently.

MSoegtropIMC:
@benediktahrens : thanks - even a somewhat blurred and possibly subjective image helps to find a way!

A note regarding the PR: As discussed I disabled unimath on 32 bit platforms. I wonder if it would make sense to note users who install the "full" platform on 32 bit Cygwin that they won't have UniMath. Most people I know using 32 bit Coq do so on 64 bit Windows simply because 32 bit Coq is a bit faster (at least on desktops cause of cache size issues, on server CPUs this is less visible).

benediktahrens:
@MSoegtropIMC : Notifying Win 32bit users would be great, possibly in the release notes. If a message can be shown during installation, that would be very helpful as well - of course, this might not be technically feasible. Thanks so much!

MSoegtropIMC:
I can notify users in the interactive installation script - which advanced users are expected to use. In the binary installers it would indeed not be easy. I need to work on proper release notes ...

PR #122, opened 2021-08-22
Avoid failing when an older version of cygwin is already installed

### LINKED ISSUE #121: coq_platform_make_windows.bat failure trying to update cygwin
It appears that my previously installed version of cygwin is no longer available as of today, which causes the build script to fail in the `source shell_scripts/init_cygwin_fixes.sh` step.  It was working yesterday.  http://mirrors.kernel.org/sourceware/cygwin/x86_64/release/mingw64-x86_64-binutils/ shows a new version that was just pushed today, Aug 22.

As a script intended to be user-friendly, there should be a message informing the user of the failure and what to do.  And then maybe have the script continue.

```
--2021-08-22 12:19:33--  http://mirrors.kernel.org/sourceware/cygwin/x86_64/release/mingw64-x86_64-binutils/mingw64-x86_64-binutils-2.35.1-1.tar.xz
Resolving mirrors.kernel.org (mirrors.kernel.org)... 2001:4f8:4:6f:0:1994:3:14, 149.20.37.36
Connecting to mirrors.kernel.org (mirrors.kernel.org)|2001:4f8:4:6f:0:1994:3:14|:80... connected.
HTTP request sent, awaiting response... 404 Not Found
2021-08-22 12:19:33 ERROR 404: Not Found.

ERROR coq_platform_make_windows.bat failed
```

jfehrle:
Fixes #121

### DIFF
--- a/coq_platform_make.sh
+++ b/coq_platform_make.sh
@@ -15,7 +15,7 @@ source shell_scripts/init_safety_debug.sh
 source shell_scripts/init_paths.sh
 source shell_scripts/init_utilities.sh
 source shell_scripts/init_machine_type.sh
-source shell_scripts/init_cygwin_fixes.sh
+source shell_scripts/init_cygwin_fixes.sh || echo; echo "*** WARNING: Unable to update cygwin, continuing ***"
 
 ###################### SETTINGS #####################
 


MSoegtropIMC:
I don't know if this is a good idea - changing the binutils has a good reason and continuing if this fails with a hardly visible warning might lead to all sorts of issues. Notably the current cygwin binutils leads to non reproducible failures of OCaml compilations (they fail randomly).

I would say the main issue is that I was busy the last few weeks and didn't have time to follow up on the CI failure reports I did receive.

MSoegtropIMC:
@jfehrle : can you do me a favor: as it looks there is a new version of binutils in Cygwin - the old one was very severely broken - about every other build failed with some obscure assembler messages in OCaml builds (something that there was unexpected stuff at the end of an ASM file). Can you please do a few local builds (3 would suffice) to check if the new binutils work? You would have to remove the binutils patch from `init_cygwin_fixes.sh`.

Your PR I will reject - it is really a bad idea to continue if these patches fail and the scripts are sufficiently complicated that it doesn't make a lot of sense to put in concrete error messages for everything. It is sufficient if maintainers know why things fail. It is a explicit design decision that the scripts rather fail before they do something possibly unreproducible.

PR #127, opened 2021-09-02
fix for gappa

gares:


gares:
snap is OK, the patch for the warning is not very nice, but I could not find better.

MSoegtropIMC:
I don't wait for the Windows CI to finish and merge right away, so that I can do a few tests.

PR #128, opened 2021-09-06
Remove open source option for CompCert

MSoegtropIMC:
since CompCert does not support it

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -24,13 +24,13 @@ on:
       platform:
         description: 'Arguments for the platform script:'
         required: true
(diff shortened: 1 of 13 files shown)


Zimmi48:
It used to be supported but it isn't anymore? What changed?

MSoegtropIMC:
I patched it myself for 3.7 - which is a bit tricky to get legally right. I asked Xavier if he wants to merge and maintain it, but he was not a big fan of the idea. And in hindsight I tend to agree with his view - CompCert maybe used for evaluation purposes which I would say includes learning VST, and it doesn't make that much sense to learn VST without ever using your own C program, for which you do need the non open source parts of CompCert anyway. So the pure open source version is of marginal use and probably not worth the effort.

PR #143, opened 2021-09-24
Adds an option for "extended picks" + various new packages

MSoegtropIMC:
It also adds various new packages to 8.13~2021.09


shell_scripts/create_smoke_test_kit.sh
```diff
@@ -248,7 +256,7 @@ echo 'ECHO "====================== SMOKE TEST SUCCESS ======================"'$'
 ##### Run bash runner script #####
 
 chmod u+x $smoke_script
-echo "On unix you can now run " $smoke_script
+echo "On macOS, linus or unix you can now run " $smoke_script
```
Zimmi48, line 259:
```suggestion
echo "On macOS, Linux or unix you can now run " $smoke_script
```

Zimmi48:
How do you decide what goes into the normal pick vs what goes into the extended pick?

MSoegtropIMC:
> How do you decide what goes into the normal pick vs what goes into the extended pick?

the criteria for full are:

- the maintainer agreed to support the package
- the package does look well maintained (I decide this by the availability date of 8.13 and 8.12 compatible tags)
- no 0.X version number
- the build time is not excessive (in case I might add a separate question as for VST)

Up to now I didn't come across anything which I thought would be controversial. The only exclusion rule I applied so far is the 0.X rule. If I come across something which is controversial, I will discuss it in the PR. Of course this can also be changed.

linux/create_snapcraft_yaml.sh
```diff
@@ -1,20 +1,40 @@
...
+# See https://creativecommons.org/publicdomain/zero/1.0/legalcode.txt
+
+###################### CREATE scnapcraft.yml #####################
+
+# This script creates snap craft yaml file from which a
+# snap based Linux installer is created in a later step
+
+###################### GATHER INFORMATION #####################
+
+# Run the Coq Platform setup scripts but stip just after opam install.
```
gares, line 18:
```suggestion
# Run the Coq Platform setup scripts but stop just after opam install.
```

coq_platform_make_windows.bat
```diff
@@ -473,6 +473,7 @@ ECHO ========== BATCH FUNCTIONS ==========
   ECHO Except for expert users this is the recommended way to run this script.
   ECHO(
   ECHO -extent=f    Setup opam and build full Coq platform
+  ECHO -extent=f    Setup opam and build extended Coq platform
```
gares, line 476:
```suggestion
  ECHO -extent=x    Setup opam and build extended Coq platform

```

shell_scripts/init_utilities.sh
```diff
@@ -112,6 +112,35 @@ function ask_user_opt3_cancel {
   done
 }
 
+# ------------------------------------------------------------------------------
+# Ask a four option + cancel question
+# Set environment variable ANSWER to the result
```
gares, line 117: an example of invokation would help

versions/packages-8.13.sh
```diff
@@ -29,71 +29,122 @@ COQ_PLATFORM_USE_DEV_REPOSITORY='N'
...
 # - Packages with system dependencies should be given first.
 #   This avoids multiple sudo password requests
 # - Packages which take a long time to build should be given last.
 #   There is some evidence that they are built early then.
 
-PACKAGES="coq.8.13.2"
+########## BASE PACKAGES ##########
+
+# Build tools - this is selected early to avoid the version is changed later
+# and everything has to be recmpiled.
```
gares, line 43:
```suggestion
# and everything has to be recompiled.
```

gares (approved):

PR #172, opened 2021-11-13
Add Publisher to Registry Entries

Trenly:
It is best practice to include the publisher name in the Add/Remove Programs table on Windows. This change sets the publisher name for the package


MSoegtropIMC (approved):

MSoegtropIMC:
Thanks for the PR - I appreciate it!

Issue #168, opened 2021-11-07
Support for Apple Silicon?


k4rtik:
Hi

When I tried the 2021.09.0 install script for macOS on an M1 Macbook Air, I received the error that `uname -m` output of "arm64" was unrecognized.

Are Apple's new SoCs supported at all?

MSoegtropIMC:
Sorry - not yet, mostly because we have no means to do CI tests.

But as far as I know opam does support it, so it shouldn't be that hard to fix. I would appreciate if you could try it.

MSoegtropIMC:
For this specific error, the fix should be to add to the case in `shell_scripts/init_machine_type.sh` these lines:

```
    arm64)
      BITSIZE=64
      HYPHEN32=
      HYPHEN64=-64
      ;;
```

You might encounter a few other places ...

k4rtik:
Ah, that's good to know. I was surprised that the lack of support was not documented anywhere in the repo.

When I get a chance to test, I will report back here.

Zimmi48:
We should probably add a link to this known issue from the release notes.

MSoegtropIMC:
@Zimmi48 : yes done. I also encouraged people who have Apple silicon to ping us for a remote debug session.

k4rtik:
Alright, with the changes that @MSoegtropIMC suggested above I got up to the following:

```
===== CHECKING VERSION OF INSTALLED OPAM =====
Found opam 2.1.0 - good!
/opt/homebrew/bin/opam
===== opam already initialized =====
===== CREATE OPAM SWITCH =====
[__coq-platform.2021.09.0.patch_ocaml] Initialised
[__coq-platform.2021.09.0.patch_coq-released] Initialised
[__coq-platform.2021.09.0.patch_coq-dev] Initialised
[coq-released] Initialised
[coq-core-dev] Initialised
[coq-extra-dev] Initialised

<><> Installing new switch packages <><><><><><><><><><><><><><><><><><><><>  🐫
Switch invariant: ["ocaml-base-compiler" {= "4.10.0"}]
[ERROR] Could not determine which packages to install for this switch:
  * Missing dependency:
    - ocaml-base-compiler = 4.10.0
    unmet availability conditions: '!(os = "macos" & arch = "arm64")'


Switch initialisation failed: clean up? ('n' will leave the switch partially installed) [Y/n] y
```

[...]

PR #181, opened 2021-11-20
coq-coqprime: Added certificate generator coq-coqprime-generator

MSoegtropIMC:


k4rtik:
Hi @MSoegtropIMC 

For the two opam files added in this PR, the same issue of CXXFLAGS vs CPPFLAGS exists. (That is, compilation fails on Apple Silicon #168)

I did not find a PR in the coq-released repo for `coq-coqprime-generator`, hence reporting it to you here and not sending a PR as this is a small change.

MSoegtropIMC:
@k4rtik : thanks! Time to grep my disk for CXXFLAGS ... I will fix it.

PR #183, opened 2021-11-22
Map the package options with descriptions

k4rtik:
If there is a better way to do this, please feel free to edit. I was wrongly under the impression that full > extended having not read the description very carefully and as the order of these options' descriptions did not match the options presented.

### DIFF
--- a/shell_scripts/ask_introduction.sh
+++ b/shell_scripts/ask_introduction.sh
@@ -18,10 +18,10 @@ cat <<EOH
 ====================== JUST COQ OR COMPLETE PLATFORM ? =======================
 This script installs the Coq Platform release $COQ_PLATFORM_RELEASE, that is:
 
-- the Coq compiler and Coq's standard library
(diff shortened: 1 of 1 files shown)


MSoegtropIMC:
@k4rtik : I appreciate your focus on detail, but it would help if you would bundle your changes into larger PRs - still separate commits but one PR. The Coq Platform CI is extremely heavy and has to run twice for every PR. There is only a limited number of runners, so that stuff I need gets queued. I will cancel the CI for this PR (it is just text).

MSoegtropIMC (approved):

k4rtik:
Sure, I wish there was an option to select whether to run the CI or not while creating the PR. I will try using the draft PR option or just create an issue instead next time.

MSoegtropIMC:
I would suggest that for small doc changes you just wait a day or two with a PR in case you find more. If not, just do the PR the usual way - I can cancel the PR CI runs when they are in the way, but the merge CI has to run.

PR #195, opened 2022-01-20
Prepare release 2022.01 (step 1)

MSoegtropIMC:


MSoegtropIMC:
The snap error seems to be a network error - merging anyway.

PR #206, opened 2022-01-27
[snap] do not purge lib/coq-core

gares:
Coq 8.15 installs stuff in lib/coq and lib/coq-core (and maybe also lib/coq-stlib).


MSoegtropIMC:
@gares : thanks! Do you think there is a need to move the tag or do we just create the snap from tag+1?

gares:
I think we can just upload the snap from tag+1 (or +2, let's see ;-)). After all it's a beta

MSoegtropIMC:
No, the snap should be from 8.1.4.1 - this is final. 8.15.0 is beta.

Zimmi48:
Still, I think this is OK that the Snap does not reflect perfectly what is tagged. I'm uneasy about moving existing tags on public repos, especially when this isn't done very briefly after the initial push.

gares:
> snap should be from 8.14.1

Hum @Zimmi48 did 8.14 had coq-core already? I'm a bit confused... (the .cmxs for the plugins are not there, so the problem is that one, but I thought that the split was done in 8.15 only).

Zimmi48:
Yes, it is Coq 8.14 that introduces coq-core.

gares:
I won't have time to test this until Monday

MSoegtropIMC:
That's fine - maybe I have time to test over the weekend.

I will fix the issue in the dev pick then.

gares:
Unexpectedly, I could test it and it works. I'll try to add a CI job which runs the smoke test in another PR.

@MSoegtropIMC if you want to see this in the store, you need to run a workflow by hand as usual setting the upload parameter to true.

MSoegtropIMC:
@gares: Thanks! I guess I can manage the store part.

MSoegtropIMC:
@gares: btw.: if you don't wait for the PR CI before merging, can you please cancel the PR runs? Coq Platform CI is quite heavy.

PR #209, opened 2022-01-30
install_opam.sh: do not enable switch by default

Blaisorblade:
Necessary (but likely insufficient) changes to avoid changing opam switch by default.

Yet untested.

### DIFF
--- a/shell_scripts/install_opam.sh
+++ b/shell_scripts/install_opam.sh
@@ -156,7 +156,7 @@ then
   COQ_PLATFORM_OPAM_MAIN_REPOS="coq-core-dev,coq-extra-dev,${COQ_PLATFORM_OPAM_MAIN_REPOS}"
 fi
 
-if ! opam switch $COQ_PLATFORM_SWITCH_NAME 2>/dev/null
+if ! opam switch list | grep -q $COQ_PLATFORM_SWITCH_NAME 2>/dev/null
 then
   echo "===== CREATE OPAM SWITCH ====="
   if [[ "$OSTYPE" == cygwin ]]
@@ -181,7 +181,7 @@ then
   $COQ_PLATFORM_TIME opam repo add --dont-select coq-extra-dev "https://coq.inria.fr/opam/extra-dev"
 
   # Create switch with the patch repo registered right away in case we need to patch OCaml
-  $COQ_PLATFORM_TIME opam switch create $COQ_PLATFORM_SWITCH_NAME $COQ_PLATFORM_OCAML_VERSION --repositories="${COQ_PLATFORM_OPAM_PATCH_REPOS},${COQ_PLATFORM_OPAM_MAIN_REPOS}"
+  $COQ_PLATFORM_TIME opam switch create --no-switch $COQ_PLATFORM_SWITCH_NAME $COQ_PLATFORM_OCAML_VERSION --repositories="${COQ_PLATFORM_OPAM_PATCH_REPOS},${COQ_PLATFORM_OPAM_MAIN_REPOS}"
 else
   echo "===== opam switch already exists ====="
 fi
@@ -191,7 +191,7 @@ fi
 # This sets the switch only locally - in case several picks are built in parallel
 eval $(opam env --set-switch --switch ${COQ_PLATFORM_SWITCH_NAME})
 
-echo === OPAM SWITECHES ===
+echo === OPAM SWITCHES ===
 opam switch
 
 echo === OPAM REPOSITORIES ===


Blaisorblade:
The smoke test on Ubuntu failed, but it's not clear it tests the right thing.

If we decided to select the new switch if there are no other ones, that test would keep working as-is but we'd want to update the help text at the end of the run.

MSoegtropIMC:
I wonder if we shouldn't add an option to the main script to not switch the opam switch. I would think most users want it to be switched.

Blaisorblade:
That sounds reasonable. My point was mostly that `shell_scripts/closing_remarks.sh` strongly implies that the switch was not changed.
```
Use the following opam commands to switch to the Coq Platform opam switch:

    opam switch ${COQ_PLATFORM_SWITCH_NAME}
    eval \$(opam env)

Or to just use this opam switch in the current shell (not setting the default):

    eval \$(opam config env --set-switch --switch ${COQ_PLATFORM_SWITCH_NAME})
```

MSoegtropIMC:
The point of the closing remark is mostly to make opam novices aware of the concept of switches - including how to run two different coq versions in two different shells.

Blaisorblade:
> The point of the closing remark is mostly to make opam novices aware of the concept of switches - including how to run two different coq versions in two different shells.

Sure. But "Use the following opam commands to switch to the Coq Platform opam switch:" isn't a reasonable message if this was already done.

MSoegtropIMC:
Should we discuss the message or the scripts?

Do you actually have a problem with the scripts doing the switch?

I would think for most people the message is fine - it would get a bit longer to say "The script already switched and in the future you can use this command to switch".

Blaisorblade:
> Should we discuss the message or the scripts?

"Message and scripts don't match" is pretty clear to me, I'm not sure what the best fix is.

> I would think for most people the message is fine - it would get a bit longer to say "The script already switched and in the future you can use this command to switch".

But it's fine to instruct them to run commands that none of your users need to actually run? That does not seem an effective way to educate users.

It's true beginners don't want complex decisions. What about checking for existing switches?
- If there is no switch, keep the current behavior. The current instructions can in fact be omitted altogether (we can try to make them less awkward, but they're not useful when given so I don't think there's a real fix).
- If there is at least one switch, don't switch automatically and give the instructions.

Personally, I'd prefer also having a command-line options to override these defaults: these scripts are also targeted at advanced users.

MSoegtropIMC:
> "Message and scripts don't match" is pretty clear to me, I'm not sure what the best fix is.

My question was: should we rather change the script or the message?

> But it's fine to instruct them to run commands that none of your users need to actually run? 

It is not required now but frequently required later. As I mentioned the main point of this is to educate users that there are switches.

I honestly really don't see what the issue is. Issuing the commands doesn't do any harm and is a nice exercise.

The only really harm could be that some very experienced users deduce from this that opam won't switch by this script and do something which makes use of this - still a rather unlikely scenario.

So yes, the message is not the whole truth, but I still think it is a good thing to tell users at this point in time.

Blaisorblade:
> My question was: should we rather change the script or the message?

Got the question, and I don't know. In my last suggestion, it's both.

Either way, the current behavior also has two issues:
- it silently switches away from the current switch during the build; this could be fixed by switching at the end
- you worried about concurrent installations — I don't know if that's a reasonable worry, but the final switch will depend on the winner of a data race. This just shows again that the script doesn't have data to make the right choice.

Blaisorblade:
Anyway, closing this since it's clear we disagree.

Zimmi48:
FWIW I tend to agree with @Blaisorblade that the message should be more consistent with the behavior to best educate users. I'd say something like:

```
I have now created a new "opam switch" with Coq Platform XXXX. To make this switch the default, you can issue the following command:

    opam switch ${COQ_PLATFORM_SWITCH_NAME}

Then, to use it in the current shell, run:

    eval \$(opam env)

If you'd rather use it in the current shell without making it the default, you can issue the following command instead:

    eval \$(opam config env --set-switch --switch ${COQ_PLATFORM_SWITCH_NAME})

Do you want me to make this switch the default for you now? [Y/n]
```

MSoegtropIMC:
@Zimmi48: yes, likely asking the user in the end if the scripts should switch (with a command line option to give this upfront) is the clearest option.

@Blaisorblade: please note that when doing multiple builds (which is a valid use case e.g. for porting stuff from Coq X to Coq Y) the builds themselves are not affected because I set the OPAM_SWITCH variable in the local shell. Since the switch is done very early in the script it is usually the last one started which remains.

@Blaisorblade: Please feel free to reopen the PR and adjust as discussed or to create an issue instead - I will then follow @Zimmi48 's suggestion + a command line switch.

PR #211, opened 2022-02-03
merge ubuntu and snap workflows, run test kit

gares:


gares:
It seems to work now. One thing that look weird is that `coqc` seems to scan the PATH:
```
====================== Running test file coq-gappa/example_20101018.v ======================
coqc  example_20101018.v
File "./example_20101018.v", line 1, characters 0-30:
Warning: Cannot open directory /usr/games/. [cannot-open-dir,filesystem]
File "./example_20101018.v", line 1, characters 0-30:
Warning: Cannot open directory /bin/. [cannot-open-dir,filesystem]
File "./example_20101018.v", line 1, characters 0-30:
Warning: Cannot open directory /sbin/. [cannot-open-dir,filesystem]
File "./example_20101018.v", line 1, characters 0-30:
Warning: Cannot open directory /usr/sbin/. [cannot-open-dir,filesystem]
```
Which seems weird to me

gares:
https://coq.zulipchat.com/#narrow/stream/237656-Coq-devs.20.26.20plugin.20devs/topic/coqc.20scanning.20PATH

gares:
I can't reproduce the issue locally.
@MSoegtropIMC would you review this one?

.github/workflows/ubuntu.yml
```diff
@@ -83,3 +90,92 @@ jobs:
...
+
+      - name: Upload Artifact to the Snap Store
+        if: ${{ github.event.inputs.upload }}
+        uses: snapcore/action-publish@v1
+        with:
+          store_login: ${{ secrets.STORE_LOGIN }}
+          snap: ${{ steps.build.outputs.snap }}
+          release: edge
+
+  # We run this one always because the Ubuntu_platform job
```
MSoegtropIMC, line 141: @gares: I thought about disabling the dev job except for scheduled builds. What is your opinion on this?

gares: works for me, but then schedled build would likely not smoke test the snap.
I guess the best is to do both: always run the smoke step, and disable dev if not cron

MSoegtropIMC (approved):
Thanks, looks good to me!

I am not sure if I don't want snap to run also on more targets - even if we don't push it people might still download it. But this can be changes in a separate step.

gares:
It may make sense to build the snap for current and next (the beta), so that when we change current we don't have surprises

PR #218, opened 2022-02-14
Update UniMath.

maggesi:
We would like to update UniMath to the last version ([2022-02-04](https://github.com/coq/opam-coq-archive/pull/2094)) in the soon to be released version of Coq Platform (https://github.com/UniMath/UniMath/issues/1442).

This is my attempt at modifying the package pick files accordingly, but I'm unsure how is the intende usage and workflow, so please revise this PR and tell if it as expected.

### DIFF
--- a/package_picks/package-pick-8.13~2022.01.sh
+++ b/package_picks/package-pick-8.13~2022.01.sh
@@ -102,7 +102,7 @@ then
 
   # Univalent Mathematics (UniMath)
   # Note: coq-unimath requires too much memory for 32 bit architectures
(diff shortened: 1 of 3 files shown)


MSoegtropIMC:
@maggesi : thanks for the effort, but for both of us it is more efficient if you just leave a note which tag you want in https://github.com/UniMath/UniMath/issues/1442. **Please don't close the issue then!**

One of the reasons is that Coq Platform PR is quite heavy and one get's nowhere unless one piggy backs a few changes into one PR.

I will close this PR.

PR #228, opened 2022-03-22
Prepare 2022.03, step 2

MSoegtropIMC:


MSoegtropIMC:
Merging despite the CI failure - all tests did run up to the "closing remarks" - the issue is that no option for switching was given, so that the script became interactive.

Issue #164, opened 2021-11-06
macOS: give installers Coq version specific names


MSoegtropIMC:
The macOS DMG installers produced by CI have all the same DMG and app name - the name should depend on the Coq version.

tech189:
It would also be good if the Windows installers had a stable naming pattern so that package managers could run a regex on the releases page to get the latest installer!

E.g. the current release title is `2022.01.0` but the installer file name contains `2022.01` without the trailing `.0`, while older versions had different patters of no Coq version or no release version. Maybe the installer file name should leave out the Coq version altogether?

tech189:
Thanks for improving the macOS releases, should I make separate issue for the Windows releases?

MSoegtropIMC:
It wouldn't hurt, but it is not required either - it will be done anyway

MSoegtropIMC:
Ah I didn't know that there is a Scoop package for Coq Platform - maybe we should put this in the docs.

tech189:
Sure, would you like me to make a PR adding a repology badge for Scoop's package to the README or something else?

MSoegtropIMC:
Yes, but please in the Windows specific Readme. Please mention it in the few lines of introduction which list the available install methods and either create a new section "package managers" and put it there are just put the Scoop link / badge as a bullet in the install methods section.

Btw.: how does scoop does it? Download and run the installer interactively?

tech189:
Will do!

It downloads the installer and extracts it as a zip file! Then it removes installer related files and creates shims for `coqtop.exe, coqc.exe, coqchk.exe, coqidetop.opt.exe` so that they are in the PATH and can be run from anywhere. Finally it sets the environment variable `COQBIN` to the `bin` folder and creates a start menu shortcut for CoqIDE.

MSoegtropIMC:
I see. In the future we will likely ave a more modular and .msi based design.

Btw.: I now aligned the MacOS and Windows names, e.g.:
```
Coq-Platform-release-2022.03.0-version~8.15~2022.03.dmg
Coq-Platform-release-2022.03.0-version~8.15~2022.03-arch-x86_64.exe
```
One can think about replacing the `arch-x86_64` with ẁindows-x84_64` and insert an `-macos` at the end for the Mac installer. What do you think?

MSoegtropIMC:
Btw.: I don't think you should be downloading ZIP files. In the releases the files are offered unzipped (the effect of ZIP compression is marginal for installers).

[...]

PR #234, opened 2022-03-24
Add package manager installation to Windows README, fix typo

tech189:
Relates to https://github.com/coq/platform/issues/164#issuecomment-1077416199

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ The table below contains links to the README files for the supported versions
 of Coq and libraries. Each README file contains a list of included packages with
 detailed information for each package.
 
-- [Coq 8.15.0 (released Jan 2022) with the first packahe pick from Mar 2022](doc/README~8.15~2022.03.md)
+- [Coq 8.15.0 (released Jan 2022) with the first package pick from Mar 2022](doc/README~8.15~2022.03.md)
 - [Coq 8.14.1 (released Nov 2021) with the first package pick from Jan 2022](doc/README~8.14~2022.01.md)
 - [Coq 8.13.2 (released Apr 2021) with an updated package pick from Jan 2022](doc/README~8.13~2022.01.md)
 - [Coq 8.13.2 (released Apr 2021) with an updated package pick from Sep 2021](doc/README~8.13~2021.09.md)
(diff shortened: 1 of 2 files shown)


tech189:
Maybe these checks which involve building the whole of Coq for every platform shouldn't run if the PR doesn't change the CI / only changes the docs

MSoegtropIMC:
@tech189 : I get few such PRs, so it is not worth the effort. I would suggest that you just create an issue and point me to your fork branch. Most PRs I don't merge anyway but piggy back them cause of the high CI cost, so the PR mechanism is essentially useless.

MSoegtropIMC:
@tech189 : I had a look at (https://community.chocolatey.org/packages?q=Coq) - it appears that they don't use Coq Platform, and if so a rather old version, so we cannot recommend this here.

For scoop I couldn't find a package list on the link you shared, but here (https://rasa.github.io/scoop-directory/by-score.html) I see that it uses Coq Platform. But the license information is wrong. Coq Platform is not LGPL 2.1, but has various licenses (see the package lists). Can you please fix this?

tech189:
(Closing the PR so the checks do not run)

I changed the Chocolatey command to one for winget which has a package for Coq Platform (https://github.com/microsoft/winget-pkgs/blob/master/manifests/c/Coq/CoqPlatform/8.14.1/Coq.CoqPlatform.installer.yaml). Feel free to incorporate my changes to the READMEs yourself, the fork is https://github.com/tech189/platform/tree/patch-1.

As for the license for Scoop's package, since the installer includes non open source software, I'll put that first and then list some of the other licenses like so: `"LGPL-2.1-only, LGPL-3.0-or-later, MIT, GPL-2.0-only, BSD-2-Clause-FreeBSD, CECILL-C, BSD-3-Clause, CECILL-B, ..."`

For `coq-compcert` would you say that "Freeware" or "Proprietary" is an acceptable label?

Is this the correct link to `coq-compcert`'s license: https://github.com/AbsInt/CompCert/blob/master/LICENSE ?

MSoegtropIMC:
The license of coq-compcert is "Proprietary" - it is free to use only in research.

tech189:
I fixed the license in Scoop's Coq package and updated it to the latest version (https://github.com/ScoopInstaller/Main/pull/3548), so you can add my installation instructions to the Windows readme, when you've got the time.

PR #236, opened 2022-03-25
add smoke test files for ott and coq-ott

palmskog:
Here I contribute test files under license CC-0 for both the Ott tool (`ott` package) and for Ott's Coq library (`coq-ott` package); the Coq library is usable standalone.

Even though only `regexp.v` is currently used in the smoke test kit, I think it's a good idea to have a test for Ott itself in case we want to validate a patch for Windows.

`regexp.ott` should be tested as follows after installation of ~~both `coq` and~~ `ott`:
```shell
ott -o regexp.v regexp.ott
```
This should result in a file `regexp.v`. This file can be further tested as below.

`regexp.v` should be tested as follows after installation of both `coq` and `coq-ott`:
```shell
coqc regexp.v
```
Note that I removed the Ott version output from the included `regexp.v`; if you generate `regexp.v` there will be a Coq comment with the Ott version number.

### DIFF
--- a/test_files/coq-ott/regexp.v
+++ b/test_files/coq-ott/regexp.v
@@ -1,0 +1,56 @@
+Require Import Arith.
+Require Import Bool.
+Require Import List.
+Require Import Ott.ott_list_core.
+
+
+Import ListNotations.
+Section RegExp.
+Variable char : Type.
+
+Definition c : Type := char. (*r character *)
+
+Inductive r : Type :=  (*r regexp *)
+ | r_zero : r
+ | r_unit : r
+ | r_char (c5:c)
+ | r_plus (r5:r) (r':r)
+ | r_times (r5:r) (r':r)
+ | r_star (r5:r).
+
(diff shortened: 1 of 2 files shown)


MSoegtropIMC:
@palmskog : thanks - this simplifies things (at least until ott is fixed on Windows).

One question: how did you disable CI?

palmskog:
I didn't disable CI, I just opened the PR. Maybe there is some issue with CI triggering? Or CI only triggers when certain files are touched?

MSoegtropIMC:
@palmskog : I piggy backed this in #238 (where I also enable CI for coq-ott).

The test file for ott I removed again since it can be extracted from the sources as all the other smoke test kit files.

I didn't add a smoke test for ott as yet, because currently I only test only coq-xyz packages and also because I don't yet have facilities to run commands, but I will enable this some day. There are other packages which would profit from this, notably CompCert and the prime certificat egenerators. Most other non coq packages are tested by some plugin.

PR #266, opened 2022-06-22
[override] Use opam pin for package override

ejgallego:
The old method was not working in all cases for two reasons:

- when a package includes build instructions in the form of $pkg.opam
  files at the root, they must be used
- a package definition often includes more files that just `$pkg/opam`

The old behavior is preserved: when a package doesn't include a build
file in the root, `opam pin` will just update the source code, and use
the `$pkg.dev` build definition present in the repos.

Not sure if the `opam/opam-coq-archive/extra-dev` repository is needed
anymore.


shell_scripts/coq_platform_override_dev_pkg.sh
```diff
@@ -15,8 +15,7 @@ do
     pkg="${override%%=*}"
     url="${override#*=}"
     echo "Overriding $pkg.dev url with $url"
-    mkdir -p opam/opam-coq-archive/extra-dev/packages/$pkg/$pkg.dev/
-    opam show --raw $pkg.dev | tr -s '\n' ' ' | sed "s@ src: [^ ]*@ src: \"$url\"@" > opam/opam-coq-archive/extra-dev/packages/$pkg/$pkg.dev/opam
+    opam pin add -y --kind=http $pkg.dev $url
```
MSoegtropIMC, line 18: I am not sure I like the -y. Coq Platform usually has set "OPAMYES=1", but when creating a pick manually I don't want any -y, because it could mess up things - I want it to stop then if there is something to ask.

ejgallego: Ok, but be aware that `opam pin` must necessary use `-y` in many cases, so I'm not sure that'd be useful for you.

Zimmi48:
You can replace the tag name by the commit hash of the main branch.

ejgallego:
Thanks @Zimmi48 , if that's safe, that's what I'll do.

Zimmi48:
Yes, that's safe, and that can be replaced by a tag after the next Platform release.

PR #277, opened 2022-08-19
Update VST to 2.11

QinshiWang:
@MSoegtropIMC
Note that `coq-vst-zlist` should belong to the open source part.
Hopefully, I have done the right change.


MSoegtropIMC:
Please don't change existing opam files - instead add new ones. It is hard to review because github mixes up the 32 and 64 bit files.

QinshiWang:
@MSoegtropIMC Git just compares the two versions and decides to display them as a renaming rather than creating new files. So there's nothing I can do. You can click on the "..." button and click "view file" to see the new file. They are identical to the OPAM files in opam-coq-archive. So once https://github.com/coq/opam-coq-archive/pull/2260 gets merged, it should be safe to merge this PR.

MSoegtropIMC:
@QinshiWang : as I said you should not remove the 2.10 packages. In Coq Platform these might be used by other picks. I will remove them in a cleanup phase. Please only **add** versions.

QinshiWang:
@MSoegtropIMC Updated as requested.

MSoegtropIMC:
Thanks for the PR! Since the opam files are already available upstream, I won't merge and close the PR.

Issue #274, opened 2022-08-10
Tracker issue for 8.16 package tag creation


MSoegtropIMC:
This issue exists to track the issues on the package upstream repos to create tags for Coq 8.16.

This ticket is referenced in all package issues.

Discussions should take place in the individual upstream issues.

### Issue links to non github projects
- [x] https://gitlab.mpi-sws.org/iris/stdpp/-/issues/142
- [x] https://gitlab.inria.fr/coquelicot/coquelicot/-/issues/6
- [x] https://gitlab.inria.fr/flocq/flocq/-/issues/21
- [x] https://gitlab.inria.fr/coqinterval/interval/-/issues/12
- [x] https://gitlab.inria.fr/gappa/coq/-/issues/11
- [x] https://gitlab.inria.fr/gappa/gappa/-/issues/12
- [x] https://gitlab.inria.fr/fpottier/menhir/-/issues/66
- [x] itauto via email to the author
- [x] https://gitlab.mpi-sws.org/iris/iris/-/issues/474

### Updates to be done
- [x] MathComp-Analysis to 0.5.3
- [x] LibHyps https://github.com/coq/opam-coq-archive/pull/2325
- [x] Update serapi - see [Zulip](https://coq.zulipchat.com/#narrow/stream/250632-Coq-Platform-devs-.26-users/topic/Serapi.20in.20snap/near/303265009)

### Local opam packages to be removed
- [x] [mczify](https://github.com/coq/platform/tree/main/opam/opam-coq-archive/released/packages/coq-mathcomp-zify/coq-mathcomp-zify.1.2.0%2B1.12%2B8.13)
- [x] coq-hott
- [x] coq-unimath
- [x] coq-unicoq
- [x] coq-mtac2
- [x] coq-elpi
- [x] coq-serapi
- [x] coq-relation-algebra

### Issues to be fixed
- [x] Do the fixes for z3_tptp in the public repo PR and locally. 
- [x] Remove .cmt, .cnti and .h files from installers.
- [x] Change CSS width of HTML ReadMe to 60em
- [x] smoke test kit for algebra-tactics does not work with 8.16
- [x] Adjust min Coq version in CompCert opam package - see (https://github.com/AbsInt/CompCert/issues/446)
- [x] Check if elpi works without -jobs 1 (patched in the local opam repo)
- [x] Include CoqIDE on Mac Fix as opam patch - see https://github.com/coq/coq/pull/16581

### Checklist
- [x] Update release in documentation
[...]

PR #285, opened 2022-09-19
Platform for 8.16: switch to OCaml 4.14.0

Blaisorblade:
OCaml 4.14 fixes some issues with debugging and seems to not have any known regression.

I understand if it's too late to change, and feel free to close if unhelpful. But I wanted to confirm this at least passes CI: it worked locally, while it didn't with the last Coq 8.15 platform (b/c of elpi and camlp5).


MSoegtropIMC:
@Blaisorblade : thanks makes sense. I am still having a lot of issues caused by using ocamlfind in 8.16 (which has the effect that hard coded paths end up in executables and various config files). I consider merging after I fixed these (and before the beta)

MSoegtropIMC:
Tracker: https://github.com/coq/platform/issues/274

PR #288, opened 2022-09-20
Update README_macOS.md

buggymcbugfix:
Explain how to add Coq bins to $PATH, since editing $PATH in `~/.zsh_rc` does not reliably work on macOS.


MSoegtropIMC:
Thanks, makes sense. I kill CI and piggy back it in one of my branches, since Coq Platform CI is rather heavy.

One question: is it guaranteed that the /etc/paths.d folder exists?

MSoegtropIMC:
Tracker: https://github.com/coq/platform/issues/274

buggymcbugfix:
I think so, I just checked a clean install of macOS 12 and it is there as an empty directory.

MSoegtropIMC:
Thanks for double checking - I couldn't find official docs on this, but I guess we are fine and if it is not there, people have at least a starting point. As I said, I will merge via piggy back.

MSoegtropIMC:
I piggy backed your PR/commit to main and close.

PR #311, opened 2022-11-16
MacOS: fix issue with macpack during DMG installer creation

MSoegtropIMC:
Co-authored-by: Enrico Tassi <Enrico.Tassi@inria.fr>

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -93,7 +93,6 @@ jobs:
         run: |
           brew install findutils
           brew install coreutils
-          brew install python@3.8
           pip3 install macpack
 
       - name: 'Build DMG installer'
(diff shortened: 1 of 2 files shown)


MSoegtropIMC:
@gares : this is a bit weird - either the homebrew on the github runners is really old or broken or intentionally messed with, or opam does something funny here. I need a few more diagnostic runs.

MSoegtropIMC:
@gares : FYI: there is a user and a system install path for python packages - since github CI runs as admin it uses the system path. as far as I can tell this can't be queried with `pip3 site`(it gives the user path).

So I came up with this slightly contrived path expression:
```
MACPACK="$(realpath "$(pip3 show macpack | grep 'Location:' | cut -f 2 -d ' ')/$(pip3 show macpack --files | grep "^[^:]*macpack$" | sed 's/^ *//')")"
```
which works locally and in CI.

Need to clean up and then I will merge.

PR #316, opened 2022-12-05
Update README_Linux.md

jaccokrijnen:
Fix link for latest .zip release (was a 404)


MSoegtropIMC:
Sorry, this is a side effect of release preparation. The versions are already bumped but the tag is not yet done.

This should work if the readme is accessed via the release page, not via main.

PR #324, opened 2023-01-26
[snap] save logs as artifacts

gares:


gares:
The PR us fine, but the logs are silly:
```
2023-01-26 09:36:09.969 Starting Snapcraft 7.2.9
2023-01-26 09:36:09.971 lifecycle command: 'pack', arguments: Namespace(bind_ssh=False, build_for=None, debug=False, destructive_mode=False, directory=None, enable_developer_debug=False, enable_experimental_extensions=False, enable_experimental_target_arch=False, enable_experimental_ua_services=False, enable_manifest=True, http_proxy=None, https_proxy=None, manifest_image_information='{"build_url":"https://github.com/coq/platform/actions/runs/4013916395"}', output=None, provider=None, target_arch=None, ua_token=None, use_lxd=False)
2023-01-26 09:36:09.971 command: pack, arguments: Namespace(bind_ssh=False, build_for=None, debug=False, destructive_mode=False, directory=None, enable_developer_debug=False, enable_experimental_extensions=False, enable_experimental_target_arch=False, enable_experimental_ua_services=False, enable_manifest=True, http_proxy=None, https_proxy=None, manifest_image_information='{"build_url":"https://github.com/coq/platform/actions/runs/4013916395"}', output=None, provider=None, target_arch=None, ua_token=None, use_lxd=False)
```

MSoegtropIMC:
Have you seen the `use_lxd=False` ?

I guess it is best I try it locally, but I won't have time before the week end.

gares:
Locally my snapcraft is using `multipass` (instead of lxd). I did a quick try with `extent=i` and it worked.

MSoegtropIMC:
You are using Nix?

gares:
> Have you seen the `use_lxd=False` ?

I don't know what that implies. I mean, I don't know if one needs lxd in order to use lxc.

gares:
No, I'm on ubuntu. Snapcraft asked me if I wanted to setup `multipass` and I said yes. I guess I don't have lxd, and IIRC setting it up is not trivial.

MSoegtropIMC:
I am also using multipass, but doesn't it use LXD below? At least the docs say so https://multipass.run:

> Multipass uses Hyper-V on Windows, QEMU and HyperKit on macOS and **LXD on Linux** for minimal overhead and the fastest possible start time.

But looking into `apt list --installed` doesn't show a trace of LXD.

Thinking about this it might run out of memory - this would explain the failure in absence of errors. I usually build 48 threads / 96 GB RAM locally, but I tried a local build with default settings now. When Unimath is enabled, the out of memory is more obvious, but out of memory can have all sorts of funny effects.

gares:
Note that I've installed multipass via snap, which also has a lxd package. I did not have it installed either. Maybe if one installs both then multipass uses lxd instead of qemu.

MSoegtropIMC:
I have multipass installed as snap as well and I don't have a LXD snap installed. So I guess it is using Qemu.

The local build with snapcraft default memory settings (which afaik means 4GB) fails in the coq build ...

MSoegtropIMC:
@gares : I cherry-picked this into my current snap hacking branch and will merge it eventually this way - even though the log is usually not useful, it sometimes is.

PR #328, opened 2023-02-13
Test CI with self hosted macOS runner

maximedenes:


rtetley:
This PR improves the platform CI introducing three things: 
- self hosted runners (which are virtualized thanks to [CIlicon](https://github.com/traderepublic/Cilicon)
- a retry loop to avoid some sporadic errors with building the DMG
- fix some of the scripts when homebrew is the package manager

.github/workflows/macos-arm.yml
```diff
@@ -112,10 +115,14 @@ jobs:
...
       - name: 'Build DMG installer'
-        shell: bash
-        run: |
-          eval $(opam env)
-          macos/create_installer_macos.sh
+        uses: Wandalen/wretry.action@master
+        with:
+          attempt_limit: 5
+          attempt_delay: 5000
+          shell: bash
```
MSoegtropIMC, line 122: @rtetley : since my shell based fix didn't work (hdiutil somehow crashes so badly that it kills the containing shell) I tried your fix. One issue: the `shell: bash` line seems to be not supported, but I used ` Wandalen/wretry.action@v1` rather than `@master`.

rtetley: On this PR it seemed to work fine ? Does it only fail when it requires a retry ?

MSoegtropIMC: It doesn't fail, it just gives a warning.

MSoegtropIMC: Sorry, likely my bad. I thought that using abbreviated version numbers (like @v1) is a github feature, but apparently it is a feature of many github actions implemented by duplicating tags - the wretry doesn't have this. So I got some random version. I will try again wit a proper version tag.

MSoegtropIMC: I tried it again with:
```
      - name: 'Build DMG installer'
        uses: Wandalen/wretry.action@v1.3.0
        with:
          attempt_limit: 5
          attempt_delay: 5000
          shell: bash
          command: |
            eval $(opam env)
            macos/create_installer_macos.sh
```
This still gives a warning:
```
Unexpected input(s) 'shell', valid inputs are ['action', 'command', 'with', 'current_path', 'attempt_limit', 'attempt_delay']
```
As far as I can tell the tag `v1.3.0` is identical to master (I don't like to have master versions of actions in CI).

Do you have a reference which suggests that the `shell` key is allowed? It would be useful to make this explicit, but it also works without it.

rtetley: Ah no, unfortunately no :-/ I hadn't noticed the warning before... I also have it on this PR...

PR #346, opened 2023-06-05
Various fix tests

MSoegtropIMC:
- limit versions of ppx.. and sexplib in coq-serapi opam file
- enable coq-serapi in Windows CI
- enable snap in CI


MSoegtropIMC:
LXD is still broken, so close this.

PR #352, opened 2023-07-17
Windows: use python3 as default python

MSoegtropIMC:
(required since python2 was removed from cygwin)


MSoegtropIMC:
@SkySkimmer : FYI: this should fix the Windows CI issue you mentioned on Zulip Monday last week. Merge is expected today evening (after CI runs through).

PR #353, opened 2023-07-18
Add editorial board

ybertot:
Adds simple references to an editorial board, mentioned mainly in the process for inclusion of new packages.

On the other hand, the members of the editorial board are not mentioned personally.  I believe this information should appear in a different file.

### DIFF
--- a/charter.md
+++ b/charter.md
@@ -40,17 +40,23 @@ Here are the different categories of expected stakeholders for the Coq Platform:
 
 - Packages are typically included on user request via a github issue.
 
-- Requests should contain a justification. Good justifications are usage of the package in a regular course or a course with more than 25 attendees or usage of the package as prerequisite in at least three other packages or research developments by authors other than the author of the original package.
+- Requests should contain a justification. Example justifications are usage of the package in a regular course or a course with a large audience or usage of the package as prerequisite in at least three other packages or research developments by authors other than the author of the original package.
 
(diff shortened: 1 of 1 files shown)


charter.md
```diff
@@ -40,18 +40,24 @@ Here are the different categories of expected stakeholders for the Coq Platform:
 
 - Packages are typically included on user request via a github issue.
 
-- Requests should contain a justification. Good justifications are usage of the package in a regular course or a course with more than 25 attendees or usage of the package as prerequisite in at least three other packages or research developments by authors other than the author of the original package.
+- Requests should contain a justification. Example justifications are usage of the package in a regular course or a course with a large audience or usage of the package as prerequisite in at least three other packages or research developments by authors other than the author of the original package.
```
palmskog, line 43: I've always felt that if enforced harshly, this paragraph (both before or after the change) has the potential to lead to "chicken and egg" problems, where researchers feel packages need to be in the Platform before they use them, but packages need to have serious research users to become included. 

How about mentioning explicitly something about the possibility of exceptions for packages that fill a specific important niche? One example would be the (already included) [LibHyps package](https://github.com/Matafou/LibHyps).

MSoegtropIMC: Indeed "filling an important niche" is also a valid justification and one could add this as example.

ybertot: I will do it, but in another branch, to avoid triggering CI at each minute change in this documentation page.

Issue #366, opened 2023-09-26
coq_makefile broken


gares:
The issue is that `-sources-of` needs to be the very first argument of `coq_makefile` to be recognized (don't ask why).
But when run from the snap, it becomes the second argument, because the first argument is `COQBIN=..`, thus everything breaks.

CFR https://coq.zulipchat.com/#narrow/stream/237977-Coq-users/topic/Coq.20suddenly.20fails.20to.20build

PR #367, opened 2023-09-26
coq_wrapper: pass COQBIN as the last argument

gares:
closes: #366


MSoegtropIMC:
Snap run into timeout. Since I am doing a local build anyway, I merge. I will then reduce the extent of the snap build to pass more likely.

MSoegtropIMC:
@gares : interestingly the snap CI fails quite consistently after this change running out of disk (and it did work consistently before). Do you have an idea why this might be?

PR #382, opened 2023-10-24
Testing removal of gappa patch

rtetley:


MSoegtropIMC:
@MSoegtropIMC : just CCing myself here to not miss notifications.

MSoegtropIMC:
FTR: removing the patch does work for the platforms we have in CI, but not in general (some Linux platforms require this, see https://github.com/ocaml/opam-repository/pull/24793.

Issue #372, opened 2023-10-05
Tracker issue for 8.18 pick creation


MSoegtropIMC:
This issue exists to track the issues on the package upstream repos to create tags for Coq 8.18.

This ticket is referenced in all package issues.

Discussions should take place in the individual upstream issues.

### Issue links to non github projects

- [x] https://gitlab.mpi-sws.org/iris/stdpp/-/issues/198
- [x] https://gitlab.inria.fr/coquelicot/coquelicot/-/issues/9
- [x] https://gitlab.inria.fr/flocq/flocq/-/issues/23
- [x] https://gitlab.inria.fr/coqinterval/interval/-/issues/14
- [x] https://gitlab.inria.fr/gappa/coq/-/issues/13
- [x] https://gitlab.inria.fr/fpottier/menhir/-/issues/73
- [x] https://gitlab.inria.fr/fbesson/itauto/-/issues/32
- [x] https://gitlab.mpi-sws.org/iris/iris/-/issues/544
- [x] https://gitlab.inria.fr/gappa/gappa/-/issues/13

### Updates to be done

- [x] Update mathcomp-word to 3.0 - see https://github.com/coq/opam/pull/2844#issuecomment-1822778094
- [x] Update libhyps - see https://github.com/Matafou/LibHyps/issues/13

### Local opam packages to be removed
- [x] coq-unicoq.1.6+8.18
- [x] coq-math-classes.8.18.0
- [x] coq-unimath.20231010
- [x] coq-compcert.3.13.1
- [x] coq-corn.8.18.0
- [x] coq-mtac2.1.4+8.18
- [x] coq-ott.0.33 see https://github.com/ocaml/opam-repository/pull/24942

### Local opam packages to be merged upstream

- [x] coq-deriving.0.1.1
- [x] coq-extructures.0.3.1
- [x] coq-simple-io.1.8.0

### Issues to be fixed
- [x] Provide some sort of smoke test for serapi (it failed to install in a few cases)
- [x] Somehow restrict the number of picks shown - either by a fixed number or by sort index 0 (this is still OK in current release => delayed)
- [x] Smoke Test coq-libhyps
- [x] Smoke Test coq-extructures
- [x] Think about providine a dependency graph as part of the doc
- [x] Coq Ott is disabled on WIndows
[...]

PR #383, opened 2023-10-31
Updating mtac2 version

rtetley:


palmskog:
@rtetley I don't understand why you need this package in the Platform repo. I've made quite a few improvements to the canonical package that lives in the Coq opam archive: https://github.com/coq/opam/blob/master/released/packages/coq-mtac2/coq-mtac2.1.4%2B8.18/opam

rtetley:
@palmskog there is a list of packages that will be removed from the platform in the tracker issue (#372). To save time I will remove them all in one go once all is done !

rtetley:
Agreed. The pull request was created at the same time as the opam one and I did not think to update it before I merged. I'll be more careful next time :-)

rtetley:
In this instance I'll just open a PR now so to take the improvements into account !

palmskog:
No big issue here really, but maybe we could agree to have a general workflow where Platform packages are first attempted to be added to the Coq opam archive (only), and then if that process stalls after a day for some reason, e.g., due to CI failures, then you add the package here.

palmskog:
By the way @rtetley please ensure you have rights to merge PRs to the Coq opam archive. I try to review and merge fairly quickly, but I may not always be available (and others such as Enrico and Guillaume Melquiond may not be available either). Maybe easiest if you ask Enrico about it.

PR #389, opened 2023-11-14
Prepare 2023.10 2

MSoegtropIMC:
This might be final, except that we want to split this into two separate picks, one based on mathcomp 1.18 and one based on 2.1 since mathcomp-analysis and mathcomp-word are not available for mc 2.x as yet.


MSoegtropIMC:
@rtetley : FYI.

PR #396, opened 2024-01-26
Package pick 8.19

rtetley:


palmskog:
@rtetley note that the package definitions here now live in `released`, and there is a crucial fix for `coq-elpi` as seen here: https://github.com/coq/opam/pull/2939

PR #406, opened 2024-03-20
Opam: undo upstream PR https://github.com/ocaml/opam-repository/pull/25484

MSoegtropIMC:
This undoes the PR in the main opam repo which states that dune.3.7.0 (which is used by the Coq 8.17 pick) is not compatible with Windows (which is not factual).


MSoegtropIMC:
This has already been fixed in opam upstream, so I will close this without merging. Still the CI run was interesting - it showed that this was the issue.

PR #416, opened 2024-06-19
Patching latest lablgtk version

rtetley:


MSoegtropIMC:
@rtetley : FYI: required changes have been merged upstream in lablgtk, so no patch to lablgtk is required any more. Instead a patch of coqide is required. This should all be fine.

rtetley:
Noted ! I'll close this then

PR #420, opened 2024-07-08
Macos upload build on failure

MSoegtropIMC:


MSoegtropIMC:
The upload of the build folder on failure works and the artefact upload and download with the updated actions works on WIndows -> merge

PR #422, opened 2024-07-12
Fix ubuntu CI failiures

rtetley:
- Remove metacoq patch for old coq versions (8.15, 8.16)
- Upgrade elpi to be the same for both 8.15 package picks
- Add custom gappa opam file until it is fixed [upstream](https://github.com/ocaml/opam-repository/pull/25896)


MSoegtropIMC:
@rtetley : I fixed all Ubuntu issues except for the snap issue. The Zulip discussion on the topic gave me some ideas on what is going on with z3, but I need to have a look at the python file which generates the makefile to find out what is going on.

I merge anyway since it is definitely an improvement (and also makes CI much faster).

PR #430, opened 2024-09-11
Give more detail on creating customized versions

jfehrle:
WDYT?  I'm happy to revise.


jfehrle:
"main" is a poor choice of branch name!

MSoegtropIMC:
> "main" is a poor choice of branch name!

Yes - it should be reserved for "main" branches (what one used to call master branch until 4 years back). See  https://www.theserverside.com/feature/Why-GitHub-renamed-its-master-branch-to-main

PR #446, opened 2024-11-16
Give more detail on creating customized versions

jfehrle:
Identical to #430, which was on branch "main" by mistake.

I'm happy to revise.

### DIFF
--- a/doc/FAQ-customized-installers.md
+++ b/doc/FAQ-customized-installers.md
@@ -7,6 +7,57 @@ The scripts for creating binary packages and installers should be able to
 handle such variants, so that it should be easy to create a custom installer
 e.g. for a lecture.
 
+## Creating released versions of packages
+
+The CP build scripts get the source code for each package by downloading a zipped file.
+For github, zipped files are available only for _released_ versions. For example, the
+[Coq releases page](https://github.com/coq/coq/releases) lists the released versions.
+The zip files are listed under "Assets".  If your changes are not merged, then the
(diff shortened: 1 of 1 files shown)


jfehrle:
> @MSoegtropIMC I appreciate the doc changes but they need to be corrected in a few ways. E.g. it is easy to get a tar ball for any commit or tag without releasing it. One should mention how this works rather than how to make a release in github. For the opam package the main problem is that most packages are in a different repo. Only the coq packages are in the repo you mentioned. As is this is very confusing. But as I said, just remove the doc changes and keep them in the separate PR.

As I recall, it took me at least 6-8 hours and to figure out how to do this stuff with your help over a couple days.  It really shouldn't be that hard.  Changing the instructions to do something else is another project.  If you give me some guidance I'll try to revise it.  Otherwise I fumble around with stuff that's not documented, kind of randomly trying one thing after another.

For the download, if I change the url to `https://api.github.com/repos/jfehrle/coq/tarball/debug_pl_8_19` the build still seems to work even if I don't update the checksum.  If I do a `wget` with that URL and checksum it, the checksum is different.  So not sure what's happening.

For downloading in order to compute the checksum, should I recommend `wget`?  I tried using `curl` but didn't find the right parameters.

MSoegtropIMC:
@jfehrle : I can adjust your text and you can review it then. I just wanted it to be out of the Ltac2 debugger PR so that we can do this separately.

PR #471, opened 2025-05-27
Fix findlib.1.9.5: replace broken camlcity.org URL with mirror

Justme0606:
This PR updates the package `findlib.1.9.5` to use a working mirror URL instead of the original source hosted on camlcity.org, which is no longer accessible.
The archive has been mirrored here: https://github.com/ocaml/ocamlfind/archive/refs/tags/findlib-1.9.5.tar.gz
The `md5` and `sha512` checksums have been updated accordingly to match the mirrored archive.
This change is necessary to ensure the Coq Platform remains buildable without relying on an unavailable external resource.


Justme0606:
Good catch, thanks ! You're right. We’re not using the non-relocatable variant. I’ll remove it to keep the repo clean.

MSoegtropIMC:
P.S.: it wouldn't hurt to port the patch to a later version of ocamlfind. The mein reason I didn't this so far is that this needs to be properly tested in installers.

Justme0606:
@MSoegtropIMC  thanks a lot ! :)

PR #478, opened 2025-08-24
Fix 8.17 download links in opam file

jim-portegies:
The download links for some packages for versions 8.17 were incorrect, as in, the checksum did not correspond. This should fix them (links copied from the upstream opam ocaml repository, checksums checked by hand, but may want to double check of course).


MSoegtropIMC:
Thank you!

I can't say I understand why CI works anyway, but it is definitely so that we should use the released tar balls and not the auto generated ones, because the auto generated ones can change with modifications in the git server (we had this a few times).

I double checked that for coq this are the only outdated tar ball URLs (except for the Ltac2 debugger preview, which is not released and consequently doesn't have a release tar ball).

PR #479, opened 2025-08-25
Only copy files if they exist

jim-portegies:
So far, the Windows installer failed when either Coqide was not installed or when compcert was not installed. This PR aims to fix that.

### DIFF
--- a/windows/Coq.nsi
+++ b/windows/Coq.nsi
@@ -411,7 +411,8 @@ FunctionEnd
 ; if sections exist (by !ifdef <section_index_var>) to decide if the license page must be included.
 ; The section index variables are only defined after the section definitions.
 
-  !define MUI_ICON "coq.ico"
+  !system "if exist coq.ico (echo !define MUI_ICON \"coq.ico\") else (echo !define MUI_ICON \"coq-shell.ico\")"
+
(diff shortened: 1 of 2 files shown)


windows/Coq.nsi
MSoegtropIMC, line 1: @jim-portegies : I guess the two icon files are the same and coq-shell.ico always exists, so I think we should always use this one and remove the ifs.

windows/create_installer_windows.sh
MSoegtropIMC, line 1: My comment is more or less the same as above: instead of switching between various icons, we should just make the method to retrieve the icon independend of CoqIDE.

MSoegtropIMC (changes requested):
In summary, we should review the way the icon is retrieved and make it independent of CoqIDE rather than adding logic which depends on if CoqIDE is there or not.

The change for CompCert is fine, thanks!

windows/Coq.nsi
jim-portegies: I do think the icons are in fact different

windows/create_installer_windows.sh
MSoegtropIMC: Yes, please. It would be nice if installers for older version still would get the old icon.

PR #483, opened 2025-09-01
First round of package-pick updates based on maintainer feedback

Justme0606:


.github/workflows/macos.yml
```diff
@@ -51,6 +51,7 @@ jobs:
           # Keep this in sync with the Smoke test below
           - '9.0+preview~2025.08'
           - '8.20~2025.01'
+
```
MSoegtropIMC, line 54: @Justme0606 : please try to avoid such non-changes in the future.

MSoegtropIMC (approved):
LGTM - will merge now.

